PR #117, opened 2019-11-05
moved "Delete" to the end of the explorer ctx menu (#116)

### LINKED ISSUE #116: OpenShift Explorer: Project: "New Component" should be on top, "Delete" in 2nd place
Steps:

1. ASSERT: OpenShift Explorer: have a cluster with a project
1. EXEC: select project and open up ctx menu
1. EXEC: look at the entries in the ctx menu

Result:
"Delete" is in the 1st place, "New Component" is 2nd. "New Component" should be on top bcs in intellij (and also Eclipse) "New" usually is the 1st entry and bcs it is the more frequent operation
image

adietish:
Signed-off-by: Andre Dietisheim <adietish@redhat.com>

moves the "Delete" to the end of the Explorer ctx menu

fixes #116


adietish:
here's how the ctx menu looks like with this change:
image
image
image

PR #121, opened 2019-11-05
added ":" to labels, top-aligned forms for create comp wiz

adietish:
* top aligns step 2 in the "create component" wizard:
  * "binary" 
  * and "git" 
* adds ":" colons to labels

## How to test changes?
Steps:
1. ASSERT: have cluster with project
1. EXEC: pick "New Component" in ctx menu for the project
1. ASSERT: "Create Component" wizard shows up
1. EXEC: choose "Git" for "Source type"

Result:
image

Steps:
1. EXEC: choose "Binary" for "Source type"

Result:
image
In both forms ":" with the labels are missing and the rows should be top aligned, dropping the unused space that's in btw. the borders and btw. them

With this change, the forms look as follows:
image
image


adietish:
@jeffmaury please review

PR #123, opened 2019-11-05
added confirmation dialogs for "Delete" actions (#120)

### LINKED ISSUE #120: Should ask for confirmation before "Delete"
[kind/Enhancement]

Steps:
1. ASSERT: have a project 
1. EXEC: in Explorer: select project and pick "Delete" in ctx menu

Result:
Project with it's content is removed.

Expected result:
The tooling should ask the user if he's sure about deleting the project (or any other resource) since one cannot recover what has been deleted.

adietish:

Adds confirmation dialogs for all "Delete" actions (Kubernetes resources, Clusters)

fixes #120 

## How to test changes?
pick "Delete" on Project, Application, Service, Storage, Clusters, etc.


adietish:
Ideally the default button would be "No" - it currently is "Yes", but it seems that there isn't a simple way (but to provide our own dialog) to achieve this with the Messages class in IntelliJ.

PR #137, opened 2019-11-14
added ApplicationTreeModel tests asserting refresh upon config changes

adietish:
depends on #136 

Adds tests to assert the refresh in ApplicationTreeModel upon changes in (the Kube-) Configuration


adietish:
> Not sure to understand as it seems this PR contains changes from #136

@jeffmaury yes it builds on the prior PR (-commit) #136 - as stated on top 'depends on #136' and thus includes it. Now that #136 was merged this should come clean and include only 1 commit.

Issue #149, opened 2020-01-10
The connector does not honor IntelliJ proxy settings


doublefx:
[kind/bug]

## What versions of software are you using?
- Operating System: Windows 7
- IntelliJ version: 2019.3
- OpenShift version: 3.11

## What did you run exactly?

- Set IntelliJ http proxy
- Install intellij-openshift-connector 

## Actual behavior
intellij-openshift-connector is proposing to download odo but fails silently.

## Expected behavior
intellij-openshift-connector honors IntelliJ http proxy and download odo

## Any logs, error output, etc?
nope

PR #150, opened 2020-01-10
Honor IDE Proxy configuration. (#149)

doublefx:
[kind/bug]

- Operating System: Windows 7
- IntelliJ version: 2019.3
- OpenShift version: 3.11

## What did you run exactly?

- Set IntelliJ http proxy
- Install intellij-openshift-connector 

## Actual behavior
intellij-openshift-connector is proposing to download odo but fails silently.

## Expected behavior
intellij-openshift-connector honors IntelliJ http proxy and download odo

nope

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/LoginAction.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/LoginAction.java
@@ -14,11 +14,14 @@ import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.ui.Messages;
 import org.jboss.tools.intellij.openshift.tree.application.ApplicationsRootNode;
 import org.jboss.tools.intellij.openshift.ui.cluster.LoginDialog;
+import org.jboss.tools.intellij.openshift.utils.ExecHelper;
 import org.jboss.tools.intellij.openshift.utils.odo.Odo;
 import org.jboss.tools.intellij.openshift.utils.UIHelper;
+import org.jboss.tools.intellij.openshift.utils.odo.OkHttpClientProxy;
 
 import javax.swing.tree.TreePath;
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.util.concurrent.CompletableFuture;
 
 public class LoginAction extends LoggedOutClusterAction {
@@ -33,9 +36,10 @@ public class LoginAction extends LoggedOutClusterAction {
             return dialog;
             });
           if (loginDialog.isOK()) {
+            ExecHelper.setProxyEnvironmentVariables(OkHttpClientProxy.buildEnvironmentVariables(loginDialog.getClusterURL()));
             odo.login(loginDialog.getClusterURL(), loginDialog.getUserName(), loginDialog.getPassword(), loginDialog.getToken());
           }
-        } catch (IOException e) {
+        } catch (IOException | URISyntaxException e) {
           UIHelper.executeInUI(() -> Messages.showErrorDialog("Error: " + e.getLocalizedMessage(), "Login"));
         }
       });
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/ExecHelper.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/ExecHelper.java
@@ -38,6 +38,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import java.util.Map;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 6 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -190,7 +193,31 @@ private String getOdoCommand() throws IOException {
           command = ProgressManager.getInstance().run(new Task.WithResult<String, IOException>(null, "Downloading Odo", true) {
             @Override
             public String compute(@NotNull ProgressIndicator progressIndicator) throws IOException {
-              OkHttpClient client = new OkHttpClient();
+              OkHttpClient client;
```
jeffmaury: Should be refactored as a separate method.

doublefx: Good point, will do tomorrow from home with my own github user.
Also, this class deserves a nice cleanup IMO, do you mind if I do it in another commit?

jeffmaury: I prefer to keep different concerns in different PRs so I would prefer another PR.
I you use a different GitHub id, don't forget the DCO (sign) header as this is required for us to accept contributions

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -67,23 +68,22 @@
 import me.snowdrop.servicecatalog.api.model.ServiceInstance;
 import me.snowdrop.servicecatalog.api.model.ServiceInstanceFluent;
 import me.snowdrop.servicecatalog.api.model.ServiceInstanceList;
-import okhttp3.OkHttpClient;
-import okhttp3.Request;
-import okhttp3.Response;
+import okhttp3.*;
```
jeffmaury: Starred import are forbidden by our policy

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -104,6 +104,8 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import static com.intellij.openapi.util.text.StringUtil.isNotEmpty;
+import static java.net.Proxy.Type.*;
```
jeffmaury: Idem

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
doublefx: Done!

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
doublefx: Done!

doublefx:
Also, I noticed the odo commands are executed by the DefaultExecutor.
Odo will not honor the IDE proxy but eventually the HTTP_PROXY, HTTPS_PROXY, NO_PROXY, ALL_PROXY environment variables if set.
Because of the way the IDE defines the proxy, there is no way obvious to determine whether it is an http or https proxy from it, therefore no obvious way to pass them to odo as environment variables using the DefaultExecutor.execute(final CommandLine command, final Map<String, String> environment) .

src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/LoginAction.java
```diff
@@ -33,9 +36,10 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object s
             return dialog;
             });
           if (loginDialog.isOK()) {
+            ExecHelper.setProxyEnvironmentVariables(OkHttpClientProxy.buildEnvironmentVariables(loginDialog.getClusterURL()));
```
jeffmaury: No that should not be managed externally like that. The tree model has an refresh method that is called when the context is changed, that should be the starting point.

src/main/java/org/jboss/tools/intellij/openshift/utils/ExecHelper.java
```diff
@@ -47,6 +48,12 @@
 public class ExecHelper {
     private static final ScheduledExecutorService SERVICE = Executors.newScheduledThreadPool(Runtime.getRuntime().availableProcessors());
 
+    private static Map<String, String> proxyEnvironmentVariables;
```
jeffmaury: Don't like to much that ExecHelper has a state like that.

doublefx: Me neither but I can't see a simpler option, a more complex would be to pass down the proxyEnvironmentVariables to the execute methods which implies a lot of refactoring and did not want to take that path.

Do you see another way?

src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/LoginAction.java
doublefx: 👍

PR #158, opened 2020-02-05
#141: adding tests

### LINKED ISSUE #141: Add tests for debug action
[kind/Enhancement]

Add Unit test / integration tests for debug action

sbouchet:
Adding Tests units for debug feature

fixes #141

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/Odo.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/Odo.java
@@ -101,6 +101,6 @@ public interface Odo {
 
     String consoleURL(OpenShiftClient client) throws IOException;
 
-    void debug(String project, String application, String context, Integer port) throws IOException;
+    void debug(String project, String application, String component, String context, Integer port) throws IOException;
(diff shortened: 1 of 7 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/Odo.java
```diff
@@ -101,6 +101,6 @@
 
     String consoleURL(OpenShiftClient client) throws IOException;
 
-    void debug(String project, String application, String context, String component, Integer port) throws IOException;
+    void debug(String project, String application, String context, Integer port) throws IOException;
```
jeffmaury: For the uniformity of the API, I would keep this parameter as seems logical to pass it when we are debugging a component (even if the CLI implementation does not use it as Odo will deduce the component name from the path.

sbouchet: reverted. see #159 to fix uniformity with the link method.

sbouchet:
@jeffmaury can you re-review please ?

PR #163, opened 2020-02-14
#152: Added service catalog availability

### LINKED ISSUE #152: Add service catalog availability
[kind/Enhancement]

We should detect if Service Catalog is available on the cluster before certains operations (create component, create service) instead of throwing an error.
This can be done by adding on new Odo method isServiceCatalogAvailable(client) and trying to cast the OpenShiftClient to the ServiceCatalogClient and catching the exception.

## Why is this needed? 

For better user interaction

sbouchet:
Services are not always available on clusters. This feature adds some checks on the action to hide/show services related actions. 

fixes #152 

## How to test changes?
Unit tests updated to take in account the service availability on clusters.
Manual test can be performed in two clusters ( one with services, one without )

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
@@ -62,11 +62,7 @@ import io.fabric8.openshift.client.OpenShiftClient;
 import io.fabric8.openshift.client.dsl.BuildConfigResource;
 import io.fabric8.openshift.client.dsl.DeployableScalableResource;
 import me.snowdrop.servicecatalog.api.client.ServiceCatalogClient;
-import me.snowdrop.servicecatalog.api.client.internal.ClusterServicePlanResource;
 import me.snowdrop.servicecatalog.api.client.internal.ServiceInstanceResource;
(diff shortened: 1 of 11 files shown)


sbouchet:
draft because it may cause merge conflicts with #162

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -618,6 +622,20 @@ public void debug(String project, String application, String context, String com
     ExecHelper.executeWithTerminal(new File(component), false, command, "debug", "port-forward", "--local-port", port.toString());
   }
 
+  @Override
+  public boolean isServiceCatalogAvailable(OpenShiftClient client){
+    try {
+      ServiceCatalogClient sc = client.adapt(ServiceCatalogClient.class);
```
jeffmaury: Don't need to perform a request to check for service availability (which is bad as you call this method in Action.isVisible that will slow down the UI significantly). The Kubernetes/OpenShiftClient has a isAdaptableMethod(Class) that does the trick

sbouchet: done.

Issue #71, opened 2019-10-03
Use JSON structured output to get the list of service types


jeffmaury:
[kind/Enhancement]

## Which functionality do you think we should update/improve?

use structured output when retrieving the list of service types

## Why is this needed? 

For better reliability

PR #187, opened 2020-04-07
#184: update odo to 1.1.1

### LINKED ISSUE #184: Update odo to 1.1.1
[kind/Enhancement]

see https://github.com/openshift/odo/releases/tag/v1.1.1

odo released a new version with numerous improvements and also 1 breaking change.

## Why is this needed? 
in order to address issues like #176, an update is mandatory.

sbouchet:
update odo to 1.1.1 and fix service catalog listing

fixes #184

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/xml/ComponentTypesDeserializer.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/xml/ComponentTypesDeserializer.java
@@ -8,12 +8,13 @@
  * Contributors:
  * Red Hat, Inc. - initial API and implementation
  ******************************************************************************/
-package org.jboss.tools.intellij.openshift.utils.odo;
+package org.jboss.tools.intellij.openshift.utils.xml;
 
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.deser.std.StdNodeBasedDeserializer;
 import com.fasterxml.jackson.databind.type.TypeFactory;
+import org.jboss.tools.intellij.openshift.utils.odo.ComponentType;
 
 import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/xml/ComponentTypesDeserializer.java
```diff
@@ -8,12 +8,13 @@
  * Contributors:
  * Red Hat, Inc. - initial API and implementation
  ******************************************************************************/
-package org.jboss.tools.intellij.openshift.utils.odo;
+package org.jboss.tools.intellij.openshift.utils.xml;
```
jeffmaury, line 11: Should be kept in the odo package

src/main/java/org/jboss/tools/intellij/openshift/utils/xml/ServiceTemplatesDeserializer.java
```diff
@@ -0,0 +1,53 @@
...
+ * Copyright (c) 2020 Red Hat, Inc.
+ * Distributed under license by Red Hat, Inc. All rights reserved.
+ * This program is made available under the terms of the
+ * Eclipse Public License v2.0 which accompanies this distribution,
+ * and is available at http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.intellij.openshift.utils.xml;
```
jeffmaury, line 11: Should be kept in the odo package

sbouchet:
> Also I don't see the changes for the odo storage list output change
There is no "storage list" method implemented yet, only the list of storage from a component implemented directly by requesting the client:
https://github.com/redhat-developer/intellij-openshift-connector/blob/bf780604c94458fafafecb4759b0ad23d4fa8b90/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java#L458

This is working with this version with no required changes in the code.
Do you ask to change the implementation to use the "odo storage list" command instead?

sbouchet:
this should also fixes #71

PR #195, opened 2020-04-29
Prepare for 0.2.2

jeffmaury:

fixes #???


src/main/resources/META-INF/plugin.xml
```diff
@@ -12,6 +12,9 @@
     ]]></description>
 
   <change-notes><![CDATA[
+  <p><b>0.2.2</b></p>
+  <p>Switch to Odo 1.1.3</p>
+  <p>Improved synchronization while debugging</p>
```
sbouchet, line 17: there is also a bugfix about the "open dashboard" action

PR #208, opened 2020-05-27
created validate workflow

### LINKED ISSUE #207: Validate PRs against All supported IJ versions
[kind/Enhancement]

Automatic validation of PRs for all supported IJ versions

## Why is this needed? 
Align with others IJ projects using same strategy ( GitHub Actions )

sbouchet:
Set up workflow for IJ validation

fixes #207

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,4 +1,4 @@
-ideaVersion = IU-2018.2
+ideaVersion=IU-2019.1
 projectVersion=0.2.5-SNAPSHOT
 jetBrainsToken=invalid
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -1,4 +1,4 @@
-ideaVersion = IU-2018.2
+ideaVersion=IU-2019.1.4
```
jeffmaury, line 1: Why 2019.1.4 and not 2019.1 ? Because if our minimum is 2019.1.4 then we need to update plugin.xml

PR #214, opened 2020-06-30
Implement plan list for Service creation

### LINKED ISSUE #188: Implement plan list for Service creation
[kind/Enhancement]

Service creation should support multiple plans 

## Why is this needed? 

When invoked, the `odo catalog list service` command returns a JSON like :

`{
            "kind": "ServiceType",
            "apiVersion": "odo.openshift.io/v1alpha1",
            "metadata": {
                "name": "sso73-x509-postgresql-persistent",
                "creationTimestamp": null
            },
            "spec": {
                "hidden": false,
                "planList": [
                    "default"
                ]
            }
}`

Each service has a planList, but right now, when Services are created, only one plan is displayed to the user, in the creation dialog box ( the first one of the list ). 
Implementation should handle more than one plan.

jeffmaury:
Fixes #188

Allow to choose plan when creating service

Fixes #188

## How to test changes?

Use the create service wizard

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/ui/service/CreateServiceDialog.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/ui/service/CreateServiceDialog.java
@@ -52,8 +52,9 @@ public class CreateServiceDialog extends DialogWrapper {
         });
         serviceTemplatesComboBox.addItemListener(item -> templateSelected((ServiceTemplate) item.getItem()));
         serviceTemplatesComboBox.setModel(new DefaultComboBoxModel(serviceTemplates));
-        serviceTemplatesComboBox.setSelectedIndex(-1);
-        serviceTemplatesComboBox.setSelectedIndex(0);
+        if (serviceTemplates.length > 0) {
+            templateSelected(serviceTemplates[0]);
+        }
     }
 
     private void templateSelected(ServiceTemplate template) {
--- a/src/main/java/org/jboss/tools/intellij/openshift/ui/service/CreateServiceDialog.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/ui/service/CreateServiceDialog.java
@@ -28,6 +28,7 @@ public class CreateServiceDialog extends DialogWrapper {
     private JTextField nameField;
     private JComboBox serviceTemplatesComboBox;
     private JTextField applicationTextField;
+    private JComboBox servicePlanComboBox;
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/jboss/tools/intellij/openshift/ui/service/CreateServiceDialog.java
```diff
@@ -45,14 +46,26 @@ public void setServiceTemplates(ServiceTemplate[] serviceTemplates) {
...
             @Override
             public Component getListCellRendererComponent(JList<?> list, Object value, int index, boolean isSelected, boolean cellHasFocus) {
-                value = ((ServiceTemplate) value).getName() + " " + ((ServiceTemplate) value).getPlan();
+                value = ((ServiceTemplate) value).getName();
                 return super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
             }
         });
+        serviceTemplatesComboBox.addItemListener(item -> templateSelected((ServiceTemplate) item.getItem()));
         serviceTemplatesComboBox.setModel(new DefaultComboBoxModel(serviceTemplates));
+        serviceTemplatesComboBox.setSelectedIndex(-1);
```
sbouchet, line 55: I'm wondering why there is this extra set selection ?

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/ServiceTemplatesDeserializer.java
```diff
@@ -34,7 +34,10 @@ public ServiceTemplatesDeserializer() {
                 for (JsonNode item : items) {
                     String name = item.get("metadata").get("name").asText();
                     //TODO manage plan lists.
```
sbouchet, line 36: please remove TODO tag

src/main/java/org/jboss/tools/intellij/openshift/ui/service/CreateServiceDialog.java
jeffmaury: That's because the first element is already the selected one so setSelectedIndex(0) did not fire the listener to be called. But I replace by a more understandable code section

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/ServiceTemplatesDeserializer.java
jeffmaury: Done in 3335bb2

PR #248, opened 2020-11-27
#246: Add support for local project devfile 

### LINKED ISSUE #246: Add support for local project devfile
[kind/Enhancement]

see also https://issues.redhat.com/browse/JBIDE-27553
and https://github.com/redhat-developer/vscode-openshift-tools/issues/1815

When a new component is created, if the selected project contains a devfile, we should disable the component type and version selection and use that local devfile.

sbouchet:

fixes #246

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/ui/component/CreateComponentModel.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/ui/component/CreateComponentModel.java
@@ -91,7 +91,7 @@ public class CreateComponentModel extends WizardModel {
 
     public void setContext(String context) {
         this.context = context;
-        File devfile = new File(context.concat(File.separator).concat(Constants.DEVFILE_NAME));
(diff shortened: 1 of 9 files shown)


src/main/java/org/jboss/tools/intellij/openshift/ui/component/CreateComponentModel.java
```diff
@@ -82,6 +91,8 @@ public String getContext() {
 
     public void setContext(String context) {
         this.context = context;
+        File devfile = new File(context.concat(File.separator).concat(Constants.DEVFILE_NAME));
```
jeffmaury, line 94: Could be rewritten as new File(context, Constants.DEVFILE_NAME)

jeffmaury (changes requested):
componentVersion should also be disabled if local project has devfile

jeffmaury (approved):

PR #255, opened 2020-12-10
Use the official oc-installer

sbouchet:
kudos to @lstocchi !

use the official Red Hat action for setting up the cluster

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,10 +34,10 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 1.8
-      - uses: manusa/actions-setup-openshift@v1.1.2
+      - uses: redhat-actions/oc-installer@v1
         name: Start cluster
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -34,10 +34,10 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 1.8
-      - uses: manusa/actions-setup-openshift@v1.1.2
+      - uses: redhat-actions/oc-installer@v1
```
jeffmaury, line 37: The reason why I choose this action is that it setup network specifically to the GitHub Actions infra so it's likely not supported by the Red Hat action 👎

jeffmaury (changes requested):

.github/workflows/ci.yml
sbouchet: yes, figured that out when workflow fails. closing :(

PR #256, opened 2020-12-11
Update to odo 2.0.3

### LINKED ISSUE #227: Switch to odo 2.0
[kind/Feature]

## Which functionality do you think we should add?

Switch to odo 2.0

## Why is this needed?

This is a major milestone and bring new features (devfile,....)

[ ] #228 
[ ] #229

sbouchet:
update odo to latest 2.0.3

fixes #227

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/actions/component/LinkServiceAction.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/actions/component/LinkServiceAction.java
@@ -15,7 +15,6 @@ import com.intellij.notification.NotificationType;
 import com.intellij.notification.Notifications;
 import com.intellij.openapi.actionSystem.AnActionEvent;
 import com.intellij.openapi.ui.Messages;
-import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.servicecatalog.api.model.ServiceInstance;
 import org.jboss.tools.intellij.openshift.KubernetesLabels;
 import org.jboss.tools.intellij.openshift.actions.OdoAction;
@@ -79,7 +78,7 @@ public class LinkServiceAction extends OdoAction {
        } else {
           UIHelper.executeInUI(() -> Messages.showWarningDialog("No services to link to", "Link service"));
         }
-      } catch (IOException | KubernetesClientException e) {
+      } catch (IOException e) {
         UIHelper.executeInUI(() -> Messages.showErrorDialog("Error: " + e.getLocalizedMessage(), "Link service"));
       }
     });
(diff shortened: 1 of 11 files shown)


sbouchet:
@jeffmaury actions are messy but tests results are expected ( 1 falling due to link service with devfile )

src/main/java/org/jboss/tools/intellij/openshift/actions/component/LinkServiceAction.java
```diff
@@ -76,7 +79,7 @@ public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object s
        } else {
           UIHelper.executeInUI(() -> Messages.showWarningDialog("No services to link to", "Link service"));
         }
-      } catch (IOException e) {
+      } catch (IOException | KubernetesClientException e) {
```
jeffmaury, line 82: Not sure why we need this exception but the patter is that the Odo interface throws IOExceptions and implementations should translate KubernetesClientException  into IOException

jeffmaury (changes requested):

src/main/java/org/jboss/tools/intellij/openshift/actions/component/LinkServiceAction.java
sbouchet: I was getting this exception when trying to link services to components and it wasn't catch. so user was facing a no op whane clicking on the menu. I'll improve to catch only IOE.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -478,9 +478,13 @@ public void logout() throws IOException {
...
-    public List<ServiceInstance> getServices(String project, String application) {
-        ServiceCatalogClient sc = client.adapt(ServiceCatalogClient.class);
-        return sc.serviceInstances().inNamespace(project).withLabelSelector(new LabelSelectorBuilder().addToMatchLabels(KubernetesLabels.APP_LABEL, application).build()).list().getItems();
+    public List<ServiceInstance> getServices(String project, String application) throws IOException {
+        try {
+            ServiceCatalogClient sc = client.adapt(ServiceCatalogClient.class);
+            return sc.serviceInstances().inNamespace(project).withLabelSelector(new LabelSelectorBuilder().addToMatchLabels(KubernetesLabels.APP_LABEL, application).build()).list().getItems();
+        } catch (KubernetesClientException e) {
+            throw new IOException("Error when getting services list: "+e.getLocalizedMessage(), e);
+        }
```
sbouchet, line 487: @jeffmaury this new implementation breaks tree node UI in https://github.com/redhat-developer/intellij-openshift-connector/blob/00a7b83a7eb03993b5d233313bf91cf071a811b0/src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationNode.java#L40

sbouchet: reverted and return empty list. in latest commit.

jeffmaury (approved):

PR #260, opened 2020-12-14
Improve UI experience

sbouchet:
- deactivate link for devfile component as it's not supported.
- added a tooltip to inform user a devfile is detected in the selected context

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,13 +15,13 @@ jobs:
 
     steps:
       - if: github.event_name == 'pull_request'
-          name: Checkout PR
-          uses: actions/checkout@v2
-          with:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
jeffmaury, line 19: Don't understand this change as workflow is also used on master

sbouchet: do not interfere when triggerred from push event. when triggered from PR, it was using the target ref ( ie master ).

sbouchet:
closing in favor of #256

PR #263, opened 2021-01-04
Add 2020.3

jeffmaury:
Use helper from intellij-common
Use 2020.3 as default

fixes #???

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ plugins {
 
 repositories {
     mavenCentral()
+    maven { url 'https://repository.jboss.org' }
 }
 
 sourceCompatibility = '1.8'
@@ -65,8 +66,9 @@ dependencies {
     compile 'org.apache.commons:commons-compress:1.19'
     compile 'org.apache.commons:commons-exec:1.3'
     compile 'org.apache.commons:commons-lang3:3.9'
-    testCompile 'org.mockito:mockito-core:2.23.4'
+    testCompile 'org.mockito:mockito-core:2.28.2'
     testCompile 'org.easytesting:fest-assert:1.4'
+    testCompile 'com.redhat.devtools.intellij:intellij-common:0.0.1-SNAPSHOT:test'
 }
 
 group 'org.jboss.tools.intellij'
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -9,6 +9,7 @@ plugins {
 
 repositories {
     mavenCentral()
+    maven { url 'https://repository.jboss.org' }
 }
 
 sourceCompatibility = '1.8'
```
sbouchet, line 15: should be 11 for being able to build.

build.gradle
```diff
@@ -65,8 +66,9 @@ dependencies {
     compile 'org.apache.commons:commons-compress:1.19'
     compile 'org.apache.commons:commons-exec:1.3'
     compile 'org.apache.commons:commons-lang3:3.9'
-    testCompile 'org.mockito:mockito-core:2.23.4'
+    testCompile 'org.mockito:mockito-core:2.28.2'
     testCompile 'org.easytesting:fest-assert:1.4'
+    testCompile 'com.redhat.devtools.intellij:intellij-common:0.0.1-SNAPSHOT:test'
```
sbouchet, line 71: locally, i cannot find the required intellij-common SNAPSHOT during the compileTestJava phase. 
I suspect that a snapshot repository has to be declared in the gradle file.

build.gradle
jeffmaury: No as we want to keep Java8 as the lower lever as only 2020.3 seems to be compiled as Java11 as the target

build.gradle
jeffmaury: A new repo has been added

sbouchet: ok, seems to be a IDEA setup issues, i had to update the project SDK to use the correct target ( 2020.3 and java11 ) to be able to run the tests.

sbouchet:
@jeffmaury : PR looks good to me now, i can build/run locally without issues. but I'm wondering why the CI still fails...

sbouchet (approved):

jeffmaury:
Yes I don't understand, my guess is that it's related to your changes to the workflow will try to revert it to see the effect

sbouchet:
> Yes I don't understand, my guess is that it's related to your changes to the workflow will try to revert it to see the effect

yes, I think it's related, too. I'm also tryingto rework the CI workflow on my side. reverting to previous working state should be fine ( where reports are uploaded and not added to the reviews checks)

sbouchet:
checks looks all good !

Issue #234, opened 2020-10-28
Remove migration of pre-odo 1.0 components


sbouchet:

sbouchet:
closed by #266

PR #265, opened 2021-01-07
Rework ci

sbouchet:
use workflow_run to publish tests reports

when pull request or push on master are performed, a new workflow is run in order to extract the test results from the integrationTest performed by the Java CI workflow. this is to workaround security issues that a new workflow with workflow_run event needs to be created.

fixes #???

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,10 @@ jobs:
         run: chmod +x gradlew
       - name: Build with Gradle
         run: ./gradlew integrationTest
+      - uses: actions/upload-artifact@v2
+        with:
+          name: linux-integration-test-reports
+          path: build/test-results
 
   build-macos:
     runs-on: macos-latest
@@ -65,7 +69,6 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Build with Gradle
-        continue-on-error: true
         run: ./gradlew build
 
   build-windows:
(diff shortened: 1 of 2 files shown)


sbouchet:
seems the intellij-common use kub client 5.0.0 and breaks the build. (https://github.com/redhat-developer/intellij-common/issues/24)

src/test/java/org/jboss/tools/intellij/openshift/utils/MessagesHelper.java
```diff
@@ -0,0 +1,49 @@
+/*******************************************************************************
```
jeffmaury, line 1: Why is this class duplicated ? That breaks the DRY rule

jeffmaury (changes requested):

src/test/java/org/jboss/tools/intellij/openshift/utils/MessagesHelper.java
sbouchet: > Why is this class duplicated ? That breaks the DRY rule

as explained, using this class from intellij-common breaks the tests compilation and the build.
moving to fabric8 implies to refactor code ( 4.12.0 -> 5.0.0 ) and could be achieve with #234 for 0.5.0 .

jeffmaury: Then we should wait until this is done otherwise we will have code duplicated and forget to remove it

jeffmaury (approved):
LGTM but I would love to see a more generic way where we could loop over the artifacts

sbouchet:
> LGTM but I would love to see a more generic way where we could loop over the artifacts

let see if the current workflow works as expected and then refactor :)

PR #268, opened 2021-01-14
fix jenkinsFile to switch to jdk11

sbouchet:

fix release build that fails due to 2020.3 requirements to use java11

fixes #???

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,7 @@
 
 node('rhel7'){
 	def javaHome = tool 'openjdk-11'
-	env.JAVA_HOME = "${mvnHome}/bin"
+	env.JAVA_HOME = "${javaHome}/bin"
 
 	stage('Checkout repo') {
 		deleteDir()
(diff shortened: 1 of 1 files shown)


sbouchet:
@jeffmaury : not sure about the jenkinsFile syntax, but jdk label should be correct.

jeffmaury (approved):

Jenkinsfile
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env groovy
 
 node('rhel7'){
-	tools {
-		jdk 'openjdk-11'
-	}
+	def javaHome = tool 'openjdk-11'
+	env.JAVA_HOME = "${mvnHome}/bin"
```
jeffmaury, line 5: Don't know why the the tools block does not work but it seems to me that `${mvnHome}` shoud be replaced by `${javaHome}`

jeffmaury (changes requested):

sbouchet:
finally get it to work. testing locally gives expected results.

PR #270, opened 2021-01-15
#266: refactor code after bumping fabric8 client version

### LINKED ISSUE #266: Support fabric8 5.0.0
[kind/Enhancement]

update fabric8 dependency to 5.0.0

## Why is this needed? 
use of redhat-developer/intellij-common requires fabric8 5.0.0

sbouchet:

refactor code and bump fabric8 client to 5.0.0

fixes #266

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoProjectDecorator.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoProjectDecorator.java
@@ -10,7 +10,6 @@
  ******************************************************************************/
 package org.jboss.tools.intellij.openshift.utils.odo;
 
-import io.fabric8.openshift.api.model.Project;
 import io.fabric8.servicecatalog.api.model.ServiceInstance;
 import org.jboss.tools.intellij.openshift.tree.application.ApplicationTreeModel;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoProjectDecorator.java
```diff
@@ -220,14 +218,10 @@ public void logout() throws IOException {
...
-                        ComponentInfo.Builder builder = new ComponentInfo.Builder();
-                        ComponentInfo info = builder.withComponentKind(kind).build();
-                        components.add(Component.of(comp.getName(), ComponentState.NOT_PUSHED, path, info));
-                    } catch (IOException e) {
-                       throw e;
-                    }
+                    ComponentKind kind = getComponentKind(path);
+                    ComponentInfo.Builder builder = new ComponentInfo.Builder();
+                    ComponentInfo info = builder.withComponentKind(kind).build();
+                    components.add(Component.of(comp.getName(), ComponentState.NOT_PUSHED, path, info));
```
adietish, line 224: these are the only lines added, they're not catched any more. The rest are removals. Can you please elaborate how these removals fix things when using the newer kubernetes-client?

sbouchet: this is an unnecessary try-catch clause. the method already throws IOE and catch clause only re-throw IOE.

adietish:
+1, works for me with the additional explanations about the code removals.

jeffmaury (approved):
LGTM

PR #275, opened 2021-01-25
Switch to intellij-common

### LINKED ISSUE #274: Refactor code and use intellij-common
[kind/Enhancement]

Lots of code is duplicated between intellij-common and intellij-openshift. We should refactor to use the code in intellij-common

## Why is this needed? 

To reduce technical debt

jeffmaury:
Fixes #274

fixes #???


sbouchet (approved):
LGTM.

PR #287, opened 2021-02-19
feat: Provides devfile code assist and validation

### LINKED ISSUE #272: Provides devfile code assist and validation
[kind/Enhancement]

As workflow is devfile based, we should provider better experience when editing devfiles.

## Why is this needed? 

JSON schema is now available: https://github.com/devfile/api/releases/download/v2.0.0/devfile.json

jeffmaury:
Fixes #272

fixes #???


sbouchet (approved):
LGTM. however i still do not understand why test is not running on 2019.2..

jeffmaury:
This is an issue in the JSON plugin: it has a packaging error so it emits an error message and the JetBrains test infra checks after each test that no error messages have been emitted otherwise test is marked failed

PR #288, opened 2021-02-22
feat: support starters

### LINKED ISSUE #247: When devfile component is created from an empty folder, handle starters
[kind/Enhancement]

When creating a component from an empty folder, we should look if the devfile has starters and if:
- if there is a single one ask to use or not starter
- if there are several starter, ask to user starter and user should select which one.

## Why is this needed? 

For better bootstrapping

sbouchet:
propose starters when the selected context is empty

fixes #247 

## How to test changes?
Create an empty folder and select it when creating a new component.

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
@@ -426,7 +426,7 @@ public class OdoCli implements Odo {
     public List<ServiceInstance> getServices(String project, String application) {
         try {
             ServiceCatalogClient sc = client.adapt(ServiceCatalogClient.class);
-             return sc.serviceInstances().inNamespace(project).withLabelSelector(new LabelSelectorBuilder().addToMatchLabels(KubernetesLabels.APP_LABEL, application).build()).list().getItems();
+            return sc.serviceInstances().inNamespace(project).withLabelSelector(new LabelSelectorBuilder().addToMatchLabels(KubernetesLabels.APP_LABEL, application).build()).list().getItems();
         } catch (KubernetesClientException e) {
             return Collections.emptyList();
         }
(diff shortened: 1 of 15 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -432,6 +426,8 @@ public void logout() throws IOException {
     public List<ServiceInstance> getServices(String project, String application) {
         try {
             ServiceCatalogClient sc = client.adapt(ServiceCatalogClient.class);
+            client.services().inNamespace(project).withLabelSelector(new LabelSelectorBuilder().addToMatchLabels(KubernetesLabels.APP_LABEL, application).build()).list().getItems();
```
jeffmaury, line 429: Not sure to understand what is the purpose of this call ?

jeffmaury (changes requested):
Seems to me that:
- as soon as a component type has a starter then it is selected whatever the content of the selected module is
- then there is no way to unselect it
openshift

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
sbouchet: good catch. i'll remove this.

sbouchet:
@jeffmaury : i've fixed the issue when switching context with not empty folder. can you review again ?

jeffmaury (changes requested):
Better but when an empty project is selected, the first starter is automatically selected and there is no way to unselect it. In JBoss Tools we display a message to the user that he can use starters but by default none is selected without user action

jeffmaury (approved):
LGTM. The only issue (but we have it also in JBT) is that once user has selected a starter, there is no way to reset to no starter

sbouchet:
> LGTM. The only issue (but we have it also in JBT) is that once user has selected a starter, there is no way to reset to no starter

filled #293 to implement it.

PR #301, opened 2021-03-25
chore: add mailer on build failures

sbouchet:

fixes #???

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,55 +1,62 @@
 #!/usr/bin/env groovy
 
 node('rhel7'){
+	def recipientList = 'sbouchet@redhat.com, jmaury@redhat.com'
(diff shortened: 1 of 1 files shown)


jeffmaury (approved):
Wouldn't it be better to send to jbosstools-builds instead of us ? Also I think that needs to be replicated on every repo external to jbosstools (IntelliJ plugins, VSCode extensions)

sbouchet:
> Wouldn't it be better to send to jbosstools-builds instead of us ? Also I think that needs to be replicated on every repo external to jbosstools (IntelliJ plugins, VSCode extensions)

i was wondering the same for recipients... i'll change it to the jboss list.
and yes, it needs to be replicated on every jenkinsFIle on every adapter job that we take care of.

PR #302, opened 2021-03-26
fix: Paste credentials from OCP4 does not work

### LINKED ISSUE #273: Paste credentials from OCP4 does not work
[kind/bug]

- Operating System: Win10Pro
- IntelliJ version: 2020.3
- OpenShift version: 0.4.0

## What did you run exactly?

Logged on developer sandbox, copied the login command to the clipboard and tried to paste it in IntelliJ. The generated token is sha256 because the new token value has now the sha256~ prefix

## Actual behavior

Can't login and token is wrongly generated

## Expected behavior

Token should be correct

jeffmaury:
Fixes #273

fixes #???


sbouchet (approved):
LGTM

PR #304, opened 2021-04-09
feat: notify telemetry of all actions (#298)

### LINKED ISSUE #298: Add telemetry info
[kind/Enhancement]

To align with other Red Hat based tooling

sbouchet:

Add telemetry on all actions 

fixes #298

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 
 plugins {
-    id "org.jetbrains.intellij" version "0.4.10"
+    id "org.jetbrains.intellij" version "0.7.2"
     id "com.adarshr.test-logger" version "1.7.0"
     id "idea"
     id "java"
@@ -20,9 +20,9 @@ intellij {
     version ideaVersion //for a full list of IntelliJ IDEA releases please see https://www.jetbrains.com/intellij-repository/releases
     pluginName 'org.jboss.tools.intellij.openshift'
     if (ideaVersion.substring(3) != "2019.1") {
-        plugins 'java', 'terminal', 'JavaScriptDebugger', 'JavaScriptLanguage', 'CSS', 'yaml'
+        plugins 'java', 'terminal', 'JavaScriptDebugger', 'JavaScriptLanguage', 'CSS', 'yaml', 'com.redhat.devtools.intellij.telemetry:0.0.1.9'
     } else {
-        plugins 'terminal', 'JavaScriptDebugger', 'JavaScriptLanguage', 'CSS', 'yaml'
+        plugins 'terminal', 'JavaScriptDebugger', 'JavaScriptLanguage', 'CSS', 'yaml', 'com.redhat.devtools.intellij.telemetry:0.0.1.9'
     }
 	updateSinceUntilBuild false
 }
@@ -53,6 +53,7 @@ sourceSets {
 }
 
 task integrationTest(type: Test) {
+    systemProperties['com.redhat.devtools.intellij.telemetry.mode'] = 'debug'
     description = 'Runs the integration tests.'
     group = 'verification'
     testClassesDirs = sourceSets.integrationTest.output.classesDirs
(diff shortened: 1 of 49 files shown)


src/main/java/org/jboss/tools/intellij/openshift/actions/application/DeleteApplicationAction.java
```diff
@@ -24,27 +24,37 @@
...
-    if (Messages.NO == Messages.showYesNoDialog("Delete Application '" + applicationNode.getName() + "'.\nAre you sure?", "Delete Application",
-        Messages.getQuestionIcon())) {
-        return;
+
+    public DeleteApplicationAction() {
+        super(ApplicationNode.class);
+    }
+
+    @Override
+    protected String getTelemetryActionName() { return "delete application"; }
```
jeffmaury, line 36: Please use a prefix. IJ Tekton uses 'ui ', better check with @adietish

jeffmaury (changes requested):
This PR is almost impossible to review because of:
- changes not related to this PR
- reformatting

src/main/java/org/jboss/tools/intellij/openshift/actions/application/DeleteApplicationAction.java
sbouchet: @jeffmaury i've introduced a prefix constant, and choose "ui/", because the action nanmes can contains spaces and "ui-" will look like "ui-delete application". with my proposal, it will be "ui/delete application". @adietish WDYT ? is "/" supported in actions names ?

build.gradle
```diff
@@ -20,9 +20,9 @@ intellij {
     version ideaVersion //for a full list of IntelliJ IDEA releases please see https://www.jetbrains.com/intellij-repository/releases
     pluginName 'org.jboss.tools.intellij.openshift'
     if (ideaVersion.substring(3) != "2019.1") {
-        plugins 'java', 'terminal', 'JavaScriptDebugger', 'JavaScriptLanguage', 'CSS', 'yaml'
```
jeffmaury, line 23: I think this can be simplified as we now always requires the java plugin as we are now 2019.3+

jeffmaury (changes requested):

src/main/java/org/jboss/tools/intellij/openshift/actions/application/DeleteApplicationAction.java
adietish: @sbouchet according to woopra "/" would have the same problem as ":": no schema would be generated automatically. The allowed separator-characters are 
* " "
* "-"
* "_"

https://github.com/redhat-developer/intellij-tekton/pull/341#issuecomment-815179517
> When a new event without a schema comes into Woopra, the system will auto-generate a basic schema so that the event can be found in reports and searched for in the Woopra interface. However, for Schemas to be auto-generated, the allowed characters are -- lower case letters, numbers, underscore, dashes, and spaces. Since the actions contain a colon ":" they'll just have to be added to the schema manually since the character will prevent it from being automatically added.

To be fully sure you can send test events and then verify in woopra that schema are generated. I'd guess they're not given what we've been told.

src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
```diff
@@ -0,0 +1,53 @@
...
+import com.redhat.devtools.intellij.telemetry.core.util.Lazy;
+
+public class TelemetryService {
+
+    public enum TelemetryResult {
+        SUCCESS, ERROR, ABORTED
+    }
+
+    public static final String VALUE_ABORTED = "aborted";
+    public static final String PREFIX_ACTION = "ui/";
```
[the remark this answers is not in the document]
adietish: as reported by woopra **"ui/"** would seem unsupported. I'd also suggest using **"ui-" ** instead

jeffmaury (approved):
LGTM

PR #307, opened 2021-04-26
fix: odo downloaded even if found on PATH

### LINKED ISSUE #306: odo 2.0.7 found, required version is 2.0.7, do you want to download odo ?
[kind/bug]

- Operating System:
The issue occurs both on Linux and Windows 10
- IntelliJ version:
The issue occurs both on 2020.1.2 and 2021.1
- OpenShift version:
The issue occurs before attempting to connect to OpenShift.

## What did you run exactly?
The IntelliJ IDEA was run with an existing version of `odo` in the path.

## Actual behavior
A panel is opened with the title:
```
    odo tool required
```
and contents:
```
    odo 2.0.7 found, required version is 2.0.7, do you want to download odo ?
```

## Expected behavior
Since the version is the same, the user should not be prompted to download the same version.

jeffmaury:
Fixes #306

fixes #???

### DIFF
--- a/src/main/resources/tools.json
+++ b/src/main/resources/tools.json
@@ -4,7 +4,7 @@
       "version": "2.0.7",
       "versionCmd": "version --client",
       "versionExtractRegExp": "odo v(\\d+[\\.\\d+]*(-.*)?)\\s.*",
-      "versionMatchRegExpr": "0\\..*",
+      "versionMatchRegExpr": "2.0.7",
       "baseDir": "$HOME/.odo",
       "platforms": {
         "win": {


src/main/resources/tools.json
```diff
@@ -4,7 +4,7 @@
       "version": "2.0.7",
       "versionCmd": "version --client",
       "versionExtractRegExp": "odo v(\\d+[\\.\\d+]*(-.*)?)\\s.*",
-      "versionMatchRegExpr": "0\\..*",
+      "versionMatchRegExpr": "2.0.7",
```
sbouchet, line 7: not sure about this. are we need to change that on every new version of do ? shouldn't be something more like "2\\..*" for all 2.X versions

jeffmaury: No because odo does not offer backward and forward compatibility so better stick with the release the plugin was built against

sbouchet (approved):
LGTM.

PR #308, opened 2021-04-26
fix: Add hyperlinks to create project and components where content is…

### LINKED ISSUE #305: When no components/services are present for a namespace, we should provide a link node
[kind/Enhancement]

When a namespace has no odo assets, nothing is displayed. We should add a link that allow user to create odo assets

## Why is this needed? 

Because uses may think plugin is broken because nothing is displayed

jeffmaury:
… empty

Fixes #305

Needs https://github.com/redhat-developer/intellij-common/pull/69 to be merged first

fixes #???

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ plugins {
 }
 
 repositories {
+    mavenLocal()
     mavenCentral()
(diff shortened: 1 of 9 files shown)


sbouchet (approved):
LGTM.

PR #312, opened 2021-05-17
chore: Fix ZIP file name in release workflow

jeffmaury:
Fixes #311

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

## Was the change discussed in an issue?
fixes #???

PR #314, opened 2021-05-18
fix typo on release zip file

sbouchet:

fixes #???


sbouchet:
dup of #312

PR #315, opened 2021-05-18
fix: hide token chars

### LINKED ISSUE #313: Hide token and password from UI when typing
[kind/Enhancement]

when logging to cluster, the token and password fields shows clear text. that should be changed to hide the contents.

## Why is this needed? 
Security and privacy :)

sbouchet:

hide chars for token field

fixes #313


jeffmaury (approved):
LGTM

Issue #309, opened 2021-05-04
Support odo 2.2 (changes s2i support)


adietish:
Odo 2.2 is about to be released. It changes how s2i is handled. We should support this new version of odo.

jeffmaury:
See https://github.com/redhat-developer/vscode-openshift-tools/issues/2113

PR #318, opened 2021-05-26
feat: Add registry management

### LINKED ISSUE #271: Add registry management
[kind/Feature]

## Which functionality do you think we should add?

We should be able to manage devfile registries:

- add a new devfile registry
- remove an existing devfile registry
- edit an existing devfile registry

## Why is this needed?

Since we moved to devfiles

See also https://issues.redhat.com/browse/JBIDE-27610

jeffmaury:
Fixes #271

fixes #???


sbouchet (changes requested):
Tested and got some questions:
- when adding this registry (https://registry.devfile.io/), I can't get any children's. is this one supported ? 
- when creating component from empty project and want to use starter, I get the following display errors) 
Capture d’écran de 2021-05-31 15-16-07

jeffmaury:
> 
> 
> Tested and got some questions:
> 
>     * when adding this registry (https://registry.devfile.io/), I can't get any children's. is this one supported ?
> 
>     * when creating component from empty project and want to use starter, I get the following display errors)
>       Capture d’écran de 2021-05-31 15-16-07

You need a recent version of odo to support non Git based registries.
Fixes the starter display issue so we should be good

sbouchet:
> 
> You need a recent version of odo to support non Git based registries.
 ok , my PR for #309 is almost ready. i'll rebase it when this one will be merged.

> Fixes the starter display issue so we should be good
Ok, trying again

sbouchet (approved):
LGTM now that the starter label is ok.

PR #320, opened 2021-06-02
feat: Depend on Kubernetes extension for cluster managment

### LINKED ISSUE #319: Depend on Kubernetes extension for cluster managment
[kind/Enhancement]

As the Kubernetes extension from Red Hat is now available, there is no need to have our own cluster view for managing contexts. Let's remove it and only use the current context from this extension.

sbouchet:
Use the Kubernetes extension to switch context. 

fixes #319


src/main/resources/META-INF/plugin.xml
```diff
@@ -72,14 +72,14 @@
...
+  <depends>com.redhat.devtools.intellij.kubernetes</depends>
 
   <depends optional="true" config-file="withJava.xml">com.intellij.modules.java</depends>
   <depends optional="true" config-file="withJavaScriptDebugger.xml">JavaScriptDebugger</depends>
   <depends optional="true" config-file="plugin-json.xml">com.intellij.modules.json</depends>
 
   <extensions defaultExtensionNs="com.intellij">
-    <!-- Add your extensions here -->
-    <toolWindow id="OpenShift" anchor="left" factoryClass="org.jboss.tools.intellij.openshift.WindowToolFactory" icon="/META-INF/pluginIcon13x13.svg"/>
+    <toolWindow id="OpenShift" anchor="left" factoryClass="org.jboss.tools.intellij.openshift.WindowToolFactory" icon="/META-INF/OpenShiftPluginIcon13x13.svg"/>
```
jeffmaury, line 82: Not sure this renaming is worth it

jeffmaury (approved):
LGTM

src/main/resources/META-INF/plugin.xml
sbouchet: I had issues with the icons of the views when running my dev instance. both kubernetes and openshift had the same icon ( kube ). renaming it fixed it.

PR #330, opened 2021-06-28
feat: Application Explorer should use the current namespace

### LINKED ISSUE #325: Application Explorer should use the current namespace
[kind/Enhancement]

In order to align with other Kube tooling, only the current namespace should be displayed

jeffmaury:
Fixes #325

fixes #???


sbouchet (approved):
LGTM

PR #331, opened 2021-07-01
feat: Integrate alizer and improve create component workflow

### LINKED ISSUE #324: Improve user experience on the create component
[kind/Enhancement]

The create component experience

## Why is this needed? 

- application should be optional and defaulted to component name
- component name can be optional and taken from the module or source directory
- devfile should be selected by alizer
- the source directory should be defaulted to the first content folder from the first module of the project

As a result, user could launch the wizard and click finish on most projects

jeffmaury:
Fixes #324

fixes #???

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/ui/component/CreateComponentDialogStep.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/ui/component/CreateComponentDialogStep.java
@@ -281,7 +281,10 @@ public class CreateComponentDialogStep extends WizardStep<CreateComponentModel> 
             for (TreeNode nodeObject : Collections.list(rootNode.children())) {
                 DefaultMutableTreeNode node = (DefaultMutableTreeNode) nodeObject;
                 ComponentType type = (ComponentType) node.getUserObject();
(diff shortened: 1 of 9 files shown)


src/main/java/org/jboss/tools/intellij/openshift/ui/component/CreateComponentDialogStep.java
```diff
@@ -274,7 +281,7 @@ private void loadModel() {
             for (TreeNode nodeObject : Collections.list(rootNode.children())) {
                 DefaultMutableTreeNode node = (DefaultMutableTreeNode) nodeObject;
                 ComponentType type = (ComponentType) node.getUserObject();
-                if (type.getName().equals(model.getComponentTypeName())) {
+                if (type.getName().equals(model.getSelectedComponentType().getName())) {
```
sbouchet, line 284: I ran into an issue where there is multiple registry with same name, and when I want to switch from the default selected one to another. the UI still select the first one.

sbouchet (changes requested):

sbouchet (approved):
new commits fixed my issue. LGTM.

PR #365, opened 2021-12-03
feat: enable create URL for kubernetes cluster

### LINKED ISSUE #361: URL creation failed on kubernetes cluster
[kind/bug]

- Operating System: RHEL 8
- IntelliJ version: 2021.1
- OpenShift version: odo 2.4.2 on minikube 1.23.0 (kube version 1.22.1)

## What did you run exactly?
create URL on pushed component

## Actual behavior
error when creating URL ( after finish dialog ):
```
 ✗  URL creation failed:
        - host must be provided in order to create URLS of Ingress Kind
```

## Expected behavior
URL creation ok

sbouchet:

update create URL dialog to add extra host parameter when cluster is pure kubenetes.

fixes #361

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
@@ -39,6 +39,7 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 import org.apache.commons.lang3.StringUtils;
+import org.jboss.tools.intellij.openshift.Constants;
 import org.jboss.tools.intellij.openshift.KubernetesLabels;
 import org.jboss.tools.intellij.openshift.telemetry.TelemetryService;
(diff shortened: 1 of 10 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -120,15 +121,23 @@ private void reportTelemetry() {
...
+                    telemetry.error(e).send();
+                } else {
+                    telemetry.error(e.toString()).send();
+                }
             }
         }
     }
 
+    @Override
+    public boolean isOpenshift(){
```
jeffmaury, line 137: Rename to isOpenShift to respect branding

jeffmaury: And the existing method should have been updated rather the deleted and new method created, that makes readers life much easier

jeffmaury (changes requested):

PR #373, opened 2022-01-26
feat: Create the "Happy Path" test for the "Openshift Connector" plugin

Ohrimenko1988:
Create the "Happy Path" test for the "Openshift Connector" plugin

##Scenario:
 - Import sample project
 - Go to the "Openshift Connector" tab
 - Expand tree
 - Create deployment using imported sample
 - Check that component is successfully deployed

https://user-images.githubusercontent.com/22724619/151194621-ccea37e5-f64e-4f9a-ad1d-fd8cabbd29cc.mp4


jeffmaury:
Not sure why you have a bin folder if you followed instructions to load the project into IJ but those files should be removed from the PR

build.gradle
```diff
@@ -9,8 +9,12 @@ plugins {
 }
 
 repositories {
-    maven { url 'https://repository.jboss.org' }
+    mavenLocal()
     mavenCentral()
+    maven { url 'https://repository.jboss.org' }
+    maven {
+        url 'https://packages.jetbrains.team/maven/p/ij/intellij-dependencies'
```
jeffmaury, line 16: Why is this required ?

build.gradle
```diff
@@ -53,7 +57,7 @@ sourceSets {
 }
 
 task integrationTest(type: Test) {
-    systemProperties['com.redhat.devtools.intellij.telemetry.mode'] = 'debug'
```
jeffmaury, line 56: Why is this removed ?

build.gradle
```diff
@@ -96,6 +101,7 @@ runIde {
 }
 
 runIdeForUiTests {
+    systemProperty "robot-server.port", System.getProperty("robot-server.port")
```
jeffmaury, line 104: Can't we use a default value in case the system property is not set

src/it/java/org/jboss/tools/intellij/openshift/BaseTest.java
```diff
@@ -11,29 +11,69 @@
 package org.jboss.tools.intellij.openshift;
 
 import com.intellij.openapi.project.Project;
-import com.intellij.testFramework.fixtures.CodeInsightTestFixture;
```
jeffmaury, line 14: You removed all CLI test that we use to make sure new version of Odo don't break. Please restore them and use another class as parent for your new UI tests

jeffmaury (changes requested):

build.gradle
odockal: Part of setup guide: https://github.com/redhat-developer/intellij-common-ui-test-library#step-1-adding-repositories

odockal:
@jeffmaury Cannot push to this PR, do not have rights, I have opened https://github.com/redhat-developer/intellij-openshift-connector/pull/387 in favor of this PR.

PR #383, opened 2022-02-22
feat: add telemetry to Devsandbox workflow

### LINKED ISSUE #376: Add telemetry to Red Hat Developer Sandbox workflow
We should report various events to telemetry when the Red Hat Developer Sandbox workflow is processed.

Events:

- get SSO token
- each of the sandbox API calls
- cluster token retrieved
- dialog closed

sbouchet:

adding telemetry to devsandbox workflow

fixes #376

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
@@ -20,6 +20,9 @@ public class TelemetryService {
     }
 
     public static final String VALUE_ABORTED = "aborted";
+    public static final String VALUE_SUCCESS = "success";
+    public static final String VALUE_FAILURE = "failure";
+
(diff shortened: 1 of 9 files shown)


src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
```diff
@@ -34,6 +37,11 @@
     public static final String IS_OPENSHIFT = "is_openshift";
     public static final String OPENSHIFT_VERSION = "openshift_version";
 
+    public static final String DEVSANDBOX_LOGIN_DIALOG="devsandbox-login-dialog";
```
jeffmaury, line 40: Would be good to have every Red Hat Developer Sandbox related action starts with a common prefix and if I remember correctly _ is not be used (see @adietish)

jeffmaury (changes requested):

jeffmaury (approved):
LGTM

PR #420, opened 2022-06-08
fix: Show delete command for non Odo components

### LINKED ISSUE #390: Show 'Delete' command for non odo components in Application Explorer View
See

https://github.com/redhat-developer/vscode-openshift-tools/issues/2378
https://issues.redhat.com/browse/JBIDE-28366

jeffmaury:
Fixes #390

fixes #???


sbouchet (approved):
LGTM. 
however, and that's next feature, we should now add those non-odo component in the describe result of apps. ( currently only odo components are listed )

jeffmaury:
> LGTM. however, and that's next feature, we should now add those non-odo component in the describe result of apps. ( currently only odo components are listed )

Not sure to get as I think this is an Odo issue ?

sbouchet:
> > LGTM. however, and that's next feature, we should now add those non-odo component in the describe result of apps. ( currently only odo components are listed )
> 
> Not sure to get as I think this is an Odo issue ?

correct. the describe command is direct use of odo describe without any json parsing.

PR #470, opened 2022-11-09
fix: Switch to odo 3.2.0 and update telemetry

### LINKED ISSUE #469: Switch to odo 3.2.0 and update telemetry
[kind/Enhancement]

jeffmaury:
Fixes #469

fixes #???

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
@@ -28,6 +28,7 @@ import com.redhat.devtools.intellij.common.kubernetes.ClusterHelper;
 import com.redhat.devtools.intellij.common.kubernetes.ClusterInfo;
 import com.redhat.devtools.intellij.common.utils.ExecHelper;
 import com.redhat.devtools.intellij.common.utils.NetworkUtils;
+import com.redhat.devtools.intellij.telemetry.core.configuration.TelemetryConfiguration;
 import com.redhat.devtools.intellij.telemetry.core.service.TelemetryMessageBuilder;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -130,7 +131,12 @@ Map of process launched for feature (dev, debug,...) related. Key is component n
         this.client = new DefaultKubernetesClient(new ConfigBuilder().build());
         try {
             this.envVars = NetworkUtils.buildEnvironmentVariables(this.getMasterUrl().toString());
-            this.envVars.put("ODO_DISABLE_TELEMETRY", "true");
+            if (TelemetryConfiguration.getInstance().isEnabled()) {
+                this.envVars.put("ODO_TRACKING_CONSENT", "yes");
+                this.envVars.put("TELEMETRY_CALLER", "intellij");
+            } else {
+                this.envVars.put("ODO_TRACKING_CONSENT", "no");
+            }
```
adietish, line 139: What if the user chooses to enable/disable telemetry later (after `OcoCli` was instantiated)?
Isn't `"ODO_TRACKING_CONSENT"` duplicating `TelemetryConfiguration.getInstance().isEnabled()`?

jeffmaury: OdoCli is created each time a cluster is loaded so it will be reloaded next connection.
Is there a way to get notified of changes ?

adietish: Yes there's a topic available. On should be able to subscribe to the topic and receive change notifications:
https://github.com/redhat-developer/intellij-redhat-telemetry/blob/690f872c66e63e64cfad9ec56d39bce33a0a7d36/src/main/java/com/redhat/devtools/intellij/telemetry/core/configuration/TelemetryConfiguration.java#L156-L161
```
    public interface ConfigurationChangedListener {
        Topic<ConfigurationChangedListener> CONFIGURATION_CHANGED =
                Topic.create("Telemetry Configuration Changed", ConfigurationChangedListener.class);


        void configurationChanged(String property, String value);
    }
```
I'm looking for usage examples.

src/main/resources/tools.json
```diff
@@ -9,17 +9,17 @@
       "silentMode": true,
       "platforms": {
         "win": {
-          "url": "https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/odo/v3.1.0/odo-windows-amd64.exe.zip",
+          "url": "https://developers.redhat.com/content-gateway/rest/mirror/pub/openshift-v4/clients/odo/v3.2.0/odo-windows-amd64.exe.zip",
```
rm3l, line 12: I guess the versions in `tools.odo.version` and `tools.odo.versionMatchRegExpr` need to be updated to `3.2.0` too, no?

rm3l (approved):

PR #476, opened 2022-11-17
chore: replaced disposable (was: project) & removed reflective buildModel (#475)

### LINKED ISSUE #475: Remove usage of reflective buildModel and use correct disposable
related to https://github.com/redhat-developer/intellij-kubernetes/issues/498

* `WindowToolFactory.buildModel` was needed back then when our least supported version was 2018. We're now requiring at least 2021.1
see here: https://github.com/redhat-developer/intellij-common/blob/main/src/main/java/com/redhat/devtools/intellij/common/tree/StructureTreeModelFactory.java#L39
```java
try {
	Constructor<StructureTreeModel> constructor = StructureTreeModel.class.getConstructor(AbstractTreeStructure.class);
	return constructor.newInstance(structure);
} catch (NoSuchMethodException e) {
	// IC 2019.3+
	Constructor<StructureTreeModel> constructor = StructureTreeModel.class.getConstructor(AbstractTreeStructure.class, Disposable.class);
	return constructor.newInstance(structure, project);
}
```
* `StructureTreeModel` and `AsyncTreeModel` are both provided with the `project` as disposable. The [SDK documentation](https://plugins.jetbrains.com/docs/intellij/disposers.html#choosing-a-disposable-parent) states the `Project` should not be used as disposable because this causes memory leaks:

	> Even though Application and Project implement Disposable, they must NEVER be used as parent disposables in plugin code. Disposables registered using those objects as parents will not be disposed when the plugin is unloaded, leading to memory leaks.

adietish:
fixes #475

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/WindowToolFactory.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/WindowToolFactory.java
@@ -10,9 +10,7 @@
  ******************************************************************************/
 package org.jboss.tools.intellij.openshift;
 
-import com.intellij.ide.util.treeView.AbstractTreeStructure;
 import com.intellij.ide.util.treeView.NodeRenderer;
-import com.intellij.openapi.Disposable;
 import com.intellij.openapi.actionSystem.ActionPlaces;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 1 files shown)


jeffmaury (approved):
LGTM

PR #481, opened 2022-11-21
feat: Remove requirement to Log out from a cluster

### LINKED ISSUE #477: Remove requirement to Log out from a cluster
[kind/Enhancement]

User should easily switch from one cluster to another. There is no need to first log out

## Why is this needed? 

Ease of use

jeffmaury:
Fixes #477

fixes #???


sbouchet (approved):
LGTM

PR #521, opened 2023-03-24
fix: Component discovery

### LINKED ISSUE #514: Component discovery fails in 1.2.0
[kind/bug]

- Operating System:
- IntelliJ version:
- OpenShift version:

## What did you run exactly?
in 1.2.0, when creating a new component, it will not show up in tree, even after restart.

## Actual behavior
create one component,  correclty shows up. create another one, and it will not appear in tree.

## Expected behavior
two components in tree

none

sbouchet:
fix component discovery mechanism

fixes #514

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
@@ -18,6 +18,7 @@ import com.redhat.devtools.intellij.common.tree.LabelAndIconDescriptor;
 import com.redhat.devtools.intellij.common.tree.MutableModel;
 import com.redhat.devtools.intellij.common.tree.MutableModelSupport;
 import io.fabric8.kubernetes.client.KubernetesClientException;
+import org.jboss.tools.intellij.openshift.Constants;
 import org.jboss.tools.intellij.openshift.utils.odo.Binding;
 import org.jboss.tools.intellij.openshift.utils.odo.Component;
 import org.jboss.tools.intellij.openshift.utils.odo.Odo;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
```diff
@@ -213,15 +216,19 @@ private Object[] getRegistryComponentTypes(DevfileRegistryNode element) {
         List<DevfileRegistryComponentTypeNode> result = new ArrayList<>();
         try {
             element.getRoot().getOdo().getComponentTypes(element.getName()).forEach(type -> result.add(new DevfileRegistryComponentTypeNode(root, element, type)));
-        } catch (IOException e) {}
+        } catch (IOException e) {
+        }
```
adietish, line 220: Shouldn't we actu upon the exception? display that registry discovery failed?

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
```diff
@@ -213,15 +216,19 @@ private Object[] getRegistryComponentTypes(DevfileRegistryNode element) {
...
     private Object[] getRegistryComponentTypeStarters(DevfileRegistryComponentTypeNode element) {
         List<DevfileRegistryComponentTypeStarterNode> result = new ArrayList<>();
         try {
-            element.getRoot().getOdo().getComponentTypeInfo(element.getName(), element.getComponentType().getDevfileRegistry().getName()).getStarters().forEach(starter -> result.add(new DevfileRegistryComponentTypeStarterNode(element.getRoot(), element, starter)));
-        } catch (IOException e) {}
+            element.getRoot().getOdo().getComponentTypeInfo(element.getName(),
+                    element.getComponentType().getDevfileRegistry().getName()).getStarters().forEach(starter -> result.add(
+                            new DevfileRegistryComponentTypeStarterNode(element.getRoot(), element, starter)));
+        } catch (IOException e) {
+        }
```
adietish, line 231: same here, no action taken upon IOException?

adietish (approved):
Other than my nitpicking because of empty catch blocks this seems to work for me. I could add a 2nd component

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
sbouchet: done.

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
sbouchet: done.

adietish (approved):
LGTM

PR #520, opened 2023-03-23
fix: Remove/fix API marked as removal for 2023-1

sbouchet:
## What is the purpose of this change? What does it change?
Remove/fix API marked as removal for 2023-1
cleanup code
minimal supported version bump to 2021.2

## Was the change discussed in an issue?
fixes #394 
fixes #515 
fixes #519

PR #522, opened 2023-03-29
build: add latest IJ workflow

sbouchet:
fixes #???


sbouchet:
to be merged after #520

PR #545, opened 2023-05-15
chore: validated against IC-2023.2, too

adietish:

adds validation against IC-2023.2


adietish:
my bad, 2023.2 is EAP so far, stable is now 2023.1.1. We need to wait.

PR #562, opened 2023-06-21
build: Added dependency on gson (2.9.0) for UI tests

richard0202:

Added dependency on Gson library (2.9.0) which is needed for running integration UI tests.

fixes https://github.com/redhat-developer/intellij-common-ui-test-library/issues/162


sbouchet (approved):
LGTM

PR #576, opened 2023-08-24
fix: Link Service action is only visible for components with a context (#552)

### LINKED ISSUE #552: Link Service action on non local component is visible
[kind/bug]

- Operating System:
- IntelliJ version:
- OpenShift version:

## What did you run exactly?
using devsandbox, there is a modelmesh-controller component available for all. 
This component is not locally created, so there is no devfile for it.
all actions related to devfile manipulation should not be visible, like link service

## Expected behavior
Link service action should not be present

adietish:
fixes #552


sbouchet (approved):
LGTM

PR #589, opened 2023-09-29
fix: dont NPE if initializing odo fails (#536)

### LINKED ISSUE #536: java.lang.NullPointerException Odo.getMasterUrl()
ENV: 
    Fedora 35

    Intellij IDEA 222.1 (Community Edition) bld #IC-222.3739.54
    Runtime version: 17.0.3+7-b469.37 amd64
    VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o

odo is in my classpath.  Verified by
  1668 > which odo
  ~/bin/odo

Exception in plugin OpenShift Toolkit by Red Hat (1.2.0.168)
java.lang.NullPointerException: Cannot invoke "org.jboss.tools.intellij.openshift.utils.odo.Odo.getMasterUrl()" because "this.delegate" is null
	at org.jboss.tools.intellij.openshift.utils.odo.OdoProjectDecorator.getMasterUrl(OdoProjectDecorator.java:258)
	at org.jboss.tools.intellij.openshift.tree.application.ApplicationsTreeStructure.lambda$createDescriptor$14(ApplicationsTreeStructure.java:246)
	at com.redhat.devtools.intellij.common.tree.LabelAndIconDescriptor.processLabel(LabelAndIconDescriptor.java:74)
	at com.redhat.devtools.intellij.common.tree.LabelAndIconDescriptor.update(LabelAndIconDescriptor.java:64)
	at com.intellij.ide.util.treeView.PresentableNodeDescriptor.getUpdatedPresentation(PresentableNodeDescriptor.java:78)
	at com.intellij.ide.util.treeView.PresentableNodeDescriptor.update(PresentableNodeDescriptor.java:30)
	at com.intellij.ui.tree.StructureTreeModel$Node.update(StructureTreeModel.java:458)
        ...

adietish:
fixes #536

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/actions/OdoAction.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/actions/OdoAction.java
@@ -16,12 +16,17 @@ import org.jboss.tools.intellij.openshift.telemetry.TelemetryHandler;
 import org.jboss.tools.intellij.openshift.telemetry.TelemetrySender;
 import org.jboss.tools.intellij.openshift.telemetry.TelemetryService;
 import org.jboss.tools.intellij.openshift.utils.odo.Odo;
+import org.jetbrains.annotations.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.swing.tree.TreePath;
 
 import static org.jboss.tools.intellij.openshift.telemetry.TelemetryService.PREFIX_ACTION;
 
 public abstract class OdoAction extends StructureTreeAction implements TelemetryHandler {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(OdoAction.class);
 
   protected TelemetrySender telemetrySender;
 
@@ -32,14 +37,23 @@ public abstract class OdoAction extends StructureTreeAction implements Telemetry
     @Override
     public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected) {
         telemetrySender = new TelemetrySender(PREFIX_ACTION + getTelemetryActionName());
-        this.actionPerformed(anActionEvent, (Object) getElement(selected), getOdo(anActionEvent));
+        Odo odo = getOdo(anActionEvent);
+        if (odo == null) {
+          return;
+        }
+        this.actionPerformed(anActionEvent, (Object) getElement(selected), odo);
     }
 
     private Odo getOdo(AnActionEvent anActionEvent) {
-        return ActionUtils.getApplicationRootNode(anActionEvent).getOdo();
+        try {
+          return ActionUtils.getApplicationRootNode(anActionEvent).getOdo().getNow(null);
+        } catch(Exception e) {
+          LOGGER.warn("Could not get odo: " + e.getMessage(), e);
+          return null;
+        }
     }
 
-    public abstract void actionPerformed(AnActionEvent anActionEvent, Object selected, Odo odo);
+    public abstract void actionPerformed(AnActionEvent anActionEvent, Object selected, @NotNull Odo odo);
 
     protected abstract String getTelemetryActionName();
 
--- a/src/main/java/org/jboss/tools/intellij/openshift/actions/binding/DeleteBindingAction.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/actions/binding/DeleteBindingAction.java
@@ -17,6 +17,7 @@ import org.jboss.tools.intellij.openshift.actions.OdoAction;
 import org.jboss.tools.intellij.openshift.telemetry.TelemetryService;
 import org.jboss.tools.intellij.openshift.tree.application.BindingNode;
 import org.jboss.tools.intellij.openshift.utils.odo.Odo;
+import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
 
@@ -26,7 +27,7 @@ public class DeleteBindingAction extends OdoAction {
     }
 
     @Override
-    public void actionPerformed(AnActionEvent anActionEvent, Object selected, Odo odo) {
+    public void actionPerformed(AnActionEvent anActionEvent, Object selected, @NotNull Odo odo) {
         try {
             BindingNode node = (BindingNode) selected;
             if (Messages.NO == Messages.showYesNoDialog("Delete binding '" + node.getName() + "'.\nAre you sure?",
--- a/src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/AboutAction.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/AboutAction.java
@@ -17,6 +17,7 @@ import com.redhat.devtools.intellij.common.utils.UIHelper;
 import org.jboss.tools.intellij.openshift.actions.OdoAction;
 import org.jboss.tools.intellij.openshift.tree.application.ApplicationsRootNode;
 import org.jboss.tools.intellij.openshift.utils.odo.Odo;
+import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
 
@@ -32,7 +33,7 @@ public class AboutAction extends OdoAction {
   protected String getTelemetryActionName() { return "about"; }
 
   @Override
-  public void actionPerformed(AnActionEvent anActionEvent, Object selected, Odo odo) {
+  public void actionPerformed(AnActionEvent anActionEvent, Object selected, @NotNull Odo odo) {
     runWithProgress((ProgressIndicator progress) -> {
       try {
         odo.about();
--- a/src/main/java/org/jboss/tools/intellij/openshift/actions/component/ShowLogComponentAction.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/actions/component/ShowLogComponentAction.java
@@ -11,8 +11,6 @@
 package org.jboss.tools.intellij.openshift.actions.component;
 
 import com.intellij.openapi.actionSystem.AnActionEvent;
-import com.intellij.openapi.progress.ProgressIndicator;
-import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.Messages;
 import com.redhat.devtools.intellij.common.utils.UIHelper;
 import org.jboss.tools.intellij.openshift.tree.application.ComponentNode;
@@ -26,8 +24,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 
-import static org.jboss.tools.intellij.openshift.actions.ActionUtils.runWithProgress;
 import static org.jboss.tools.intellij.openshift.telemetry.TelemetryService.TelemetryResult;
 
 public class ShowLogComponentAction extends ContextAwareComponentAction {
@@ -37,7 +35,7 @@ public class ShowLogComponentAction extends ContextAwareComponentAction {
   protected String getTelemetryActionName() { return "show component log"; }
 
     protected String getActionName() {
-        return "Show log";
+        return "Show Log";
     }
 
     @Override
@@ -46,7 +44,11 @@ public class ShowLogComponentAction extends ContextAwareComponentAction {
         try {
             if (visible) {
                 ComponentNode componentNode = (ComponentNode) selected;
-                visible = isDevOrDebugAndLogNotRunning(componentNode) || isDeployAndLogNotRunning(componentNode);
+                Odo odo = componentNode.getRoot().getOdo().getNow(null);
+                if (odo == null) {
+                  return false;
+                }
+                visible = isDevOrDebugAndLogNotRunning(componentNode, odo) || isDeployAndLogNotRunning(componentNode, odo);
             }
         } catch (IOException e) {
             LOGGER.warn(e.getLocalizedMessage(), e);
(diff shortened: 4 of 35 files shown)


adietish:
@sbouchet this PR has 2 commits:
* 1st commit: simples fix
* 2nd commit: change in how odo is downloaded if it doesnt exist yet
Would love to get your take on this 2nd commit which I believe is an improvement/simplification of the whole approach

sbouchet:
<snip>
> I tried to address this as follows:
> 
> 1. get rid of explicitly calling `initializeOdo()` . Load it transparently whenever `getOdo()`  is called
> 2. get rid of the instance var `odo` and `error`. Replace by an instance var `odoFuture` which is a `CompletableFuture`. It holds odo and the possible error.
> 
> As a result of this change any time user code wants to use the odo binary it has to try/catch to get errors that may have happened when loading the binary and null-check in case the odo binary wasnt loaded yet:
> 
> ```java
>  try {
>      Odo odo = componentNode.getRoot().getOdo().getNow(null);
>      if (odo == null) {
>          return;
>      }
>      ...
>  } catch (Exception ex) {
> ```

I like the idea to just call getOdo() that returns a future. this will also simplyfy the codebase by removing the odoprojectdecorator class. +1 for implement this.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoProjectDecorator.java
```diff
@@ -1,300 +0,0 @@
...
-    }
-
-    @Override
-    public List<ComponentMetadata> analyze(String path) throws IOException {
-        return delegate.analyze(path);
-    }
-
-    @Override
-    public void createComponent(String project, String componentType, String registryName, String component, String source, String devfile, String starter) throws IOException {
-        if (StringUtils.isNotBlank(starter)) {
```
sbouchet, line 82: please move this logic in OdoCli or in the calling action

datho7561 (approved):
I can confirm that this presents the error in the OpenShift tree view when the tools.json is corrupted. The refactoring looks good as well. Thanks, Andre!

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoProjectDecorator.java
```diff
@@ -1,300 +0,0 @@
...
-    }
-
-    @Override
-    public void login(String url, String userName, char[] password, char[] token) throws IOException {
-        delegate.login(url, userName, password, token);
-    }
-
-    @Override
-    public List<Component> getComponents(String project) throws IOException {
-        List<Component> components = delegate.getComponents(project);
```
sbouchet, line 185: please move this logic i in the calling action

src/main/java/org/jboss/tools/intellij/openshift/actions/binding/DeleteBindingAction.java
```diff
@@ -26,6 +26,9 @@ public DeleteBindingAction() {
 
     @Override
     public void actionPerformed(AnActionEvent anActionEvent, Object selected, Odo odo) {
+        if (odo == null) {
```
sbouchet, line 29: not needed as already taken care of in OdoAction

src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/AboutAction.java
```diff
@@ -32,6 +32,9 @@ public AboutAction() {
 
   @Override
   public void actionPerformed(AnActionEvent anActionEvent, Object selected, Odo odo) {
+    if (odo == null) {
```
sbouchet, line 35: not needed as already taken care of in OdoAction

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsRootNode.java
```diff
@@ -74,18 +74,21 @@ public void setLogged(boolean logged) {
         this.logged = logged;
     }
 
-    public CompletableFuture<Odo> initializeOdo() {
-        return OdoCliFactory.getInstance().getOdo(project).whenComplete((odo, err) -> {
-            if (this.odo != null) {
-                this.odo.release();
-            }
-            this.odo = new OdoProjectDecorator(odo, this);
-            loadProjectModel(project);
```
sbouchet, line 83: this call is still required when loading the root node in applicationTreeStructure class

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoProjectDecorator.java
adietish: I believe that I cannot remove `OdoProjectDecorator`. Turns out that this "decorator" does things to the tree before/after invoking the odo binary. Example:
```java
    public void deleteComponent(String project, String context, String component, ComponentKind kind) throws IOException {
        if (context != null) {
            root.removeContext(new File(context));
        }
        delegate.deleteComponent(project, context, component, kind);
    }
```
I could implement the `Odo` interface in `ApplicationsRootNode` but would loose the separation of concerns that we have now. 
To me it looks as if keeping the decorator is the better option. But I'll rename it to `ApplicationsRootNodeOdo` to better reflect its purpose.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoProjectDecorator.java
adietish: I decided to keep `OdoProjectDecorator`. See arguments in comments below.

adietish:
@sbouchet corrected what you requested. Please re-review.

src/main/java/org/jboss/tools/intellij/openshift/actions/OdoAction.java
```diff
@@ -36,12 +40,32 @@ protected OdoAction(Class... filters) {
     @Override
     public void actionPerformed(AnActionEvent anActionEvent, TreePath path, Object selected) {
         telemetrySender = new TelemetrySender(PREFIX_ACTION + getTelemetryActionName());
-        this.actionPerformed(anActionEvent, (Object) getElement(selected), getOdo(anActionEvent));
+        Odo odo = getOdo(anActionEvent);
+        if (odo == null) {
+          return;
+        }
+        this.actionPerformed(anActionEvent, (Object) getElement(selected), odo);
```
sbouchet, line 47: at this point, we're sure that odo isn't null, so maybe add notNull annotation on the abstract might help for docs purpose.

src/main/java/org/jboss/tools/intellij/openshift/actions/binding/DeleteBindingAction.java
sbouchet: see above.

src/main/java/org/jboss/tools/intellij/openshift/actions/OdoAction.java
adietish: We're not sure that it isn't `null`. See:
```Java
       Odo odo = getOdo(anActionEvent);
        if (odo == null) {
          return;
        }
```
getOdo() can return `null` in many cases:

```
    private Odo getOdo(AnActionEvent anActionEvent) {
        try {
          Tree tree = getTree(anActionEvent);
          if (tree == null) {
            return null;
          }
          ApplicationsTreeStructure structure = (ApplicationsTreeStructure) tree.getClientProperty(Constants.STRUCTURE_PROPERTY);
          if (structure == null) {
            return null;
          }
          ApplicationsRootNode root = (ApplicationsRootNode) structure.getApplicationsRoot();
          if (root == null) {
            return null;
          }
          return root.getOdo().getNow(null);
        } catch(Exception e) {
          LOGGER.warn("Could not get odo: " + e.getMessage(), e);
          return null;
        }
    }
```

src/main/java/org/jboss/tools/intellij/openshift/actions/component/ShowLogComponentAction.java
```diff
@@ -123,11 +123,19 @@ private  boolean isDebug(Component component) {
     }
 
     private  boolean isLogRunningForDevOrDebug(ComponentNode componentNode) throws IOException {
-        return componentNode.getRoot().getOdo().isLogRunning(componentNode.getComponent().getPath(), componentNode.getComponent().getName(), false);
+        Odo odo = componentNode.getRoot().getOdo().getNow(null);
```
sbouchet, line 126: odo is already present before this call, better to pass it along. and it's also not null at this point ( from OdoAction )

src/main/java/org/jboss/tools/intellij/openshift/actions/component/ShowLogComponentAction.java
```diff
@@ -123,11 +123,19 @@ private  boolean isDebug(Component component) {
...
+        if (odo == null) {
+            return false;
+        }
+        return odo.isLogRunning(componentNode.getComponent().getPath(), componentNode.getComponent().getName(), false);
     }
 
     private  boolean isLogRunningForDeploy(ComponentNode componentNode) throws IOException {
-        return componentNode.getRoot().getOdo().isLogRunning(componentNode.getComponent().getPath(), componentNode.getComponent().getName(), true);
+        Odo odo = componentNode.getRoot().getOdo().getNow(null);
```
sbouchet, line 134: same as above

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationRootNodeOdo.java
```diff
@@ -80,14 +90,7 @@ public List<ComponentMetadata> analyze(String path) throws IOException {
...
         if (StringUtils.isNotBlank(starter)) {
-            File tmpdir;
-            if (SystemInfo.isWindows) {
-                tmpdir = Files.createTempDirectory("odotmp").toFile();
-            } else {
-                FileAttribute<Set<PosixFilePermission>> attr = PosixFilePermissions.asFileAttribute(
-                        PosixFilePermissions.fromString("rwxr-x---"));
-                tmpdir = Files.createTempDirectory("odotmp", attr).toFile();
-            }
+            File tmpdir = FileUtil.createTempDirectory("odotmp", null);
```
sbouchet, line 93: good catch of using this IDE util :+1:

sbouchet (commented):
please also review the sonar code smells

src/main/java/org/jboss/tools/intellij/openshift/actions/component/ShowLogComponentAction.java
adietish: can do, I agree. But these methods are also called from #isVisible where odo isnt' present. So we need to add odo retrieval there.

src/main/java/org/jboss/tools/intellij/openshift/actions/project/CreateProjectAction.java
```diff
@@ -50,14 +51,18 @@ private void doActioPerformed(ApplicationsRootNode clusterNode) {
     if ((projectName != null) && projectName.trim().length() > 0) {
       CompletableFuture.runAsync(() -> {
         try {
+          Odo odo = clusterNode.getOdo().getNow(null);
```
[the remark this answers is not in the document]
adietish: Can do, I agree. But then we also need to add odo retrieval in #execute where #doActionPerformed is called, too

adietish:
> please also review the sonar code smells

There are none left. Sonar keeps reporting 2 false positives which I need to close manually on each push.

adietish:
@sbouchet corrected things according to your comments. Please re-review.

src/main/java/org/jboss/tools/intellij/openshift/actions/OdoAction.java
sbouchet: maybe i'm wrong but 
```
        if (odo == null) {
          return;
        }
```
will break execution if `odo` is null, meaning that  on next line,

`this.actionPerformed(anActionEvent, (Object) getElement(selected), odo);`

odo isn't null for the rest of the execution ?

src/main/java/org/jboss/tools/intellij/openshift/actions/component/ShowLogComponentAction.java
sbouchet: ok then, better to pass a not null odo to those methods.

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
```diff
@@ -79,12 +77,6 @@ public ApplicationsTreeStructure(Project project) {
     }
 
     public Object getApplicationsRoot() {
-        if (!initialized.getAndSet(true)) {
```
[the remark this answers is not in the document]
adietish: I believe that this call isn't necessary any more. All calls to `getApplicationsRoot()` then get odo from it. Odo then would call `loadProjectModel` if odo isn't initialized yet.

src/main/java/org/jboss/tools/intellij/openshift/actions/component/ShowLogComponentAction.java
adietish: added the annotation on `OdoAction.actionPerformed` and all it's subclasses:

```java
public actionPerformed(AnActionEvent anActionEvent, Object selected, @NotNull Odo odo)
```

adietish:
> seems also that test coverage is missing. need to address that in another issue

I agree. Looked through my changes and all what could make sense here imho is to add a test for `ApplicationsRootNodeOdo` (added/moved in this PR) and file an issue to improve test coverage.

adietish:
@sbouchet I added a unit test for `ApplicationsRootNodeOdo`: https://github.com/redhat-developer/intellij-openshift-connector/pull/589/files#diff-148812f161fd7910c4d4c92503a07efee561cc3fc3ce795f7619aa49db4dc0fa
Please re-review. Thanks.

sbouchet (approved):
LGTM with finals changes.

PR #597, opened 2023-10-17
fix: fix build for next IJ version and fix Sonar coverage, bumping gradle

### LINKED ISSUE #591: Github action "Latest IJ" is failing
https://github.com/redhat-developer/intellij-openshift-connector/actions/runs/6450872457/job/17510762649

> com.intellij.testFramework.TestLoggerFactory$TestLoggerAssertionError: Trying to add extensions to non-registered file type Properties [Plugin: com.intellij.java]
  Caused by: com.intellij.diagnostic.PluginException: Trying to add extensions to non-registered file type Properties [Plugin: com.intellij.java]

>Expiring Daemon because JVM heap space is exhausted

sbouchet:
fix next IJ build, fix sonar/jacoco coverage, bump gradle to 8.4

fixes #591


build.gradle
```diff
@@ -136,7 +137,6 @@ dependencies {
             'io.fabric8:openshift-client:6.4.1',
             'org.apache.commons:commons-compress:1.21',
             'org.apache.commons:commons-exec:1.3',
-            'org.apache.commons:commons-lang3:3.9',
```
adietish, line 139: like dropping apache commons and use jetbrains platform instead. 👍

build.gradle
```diff
@@ -18,7 +18,8 @@ buildscript {
 }
 
 plugins {
-    id 'org.jetbrains.intellij' version '1.15.0'
+    id 'org.jetbrains.kotlin.jvm' version '1.9.10'
```
adietish, line 21: why do we need kotlin?

adietish (approved):
LGTM, maybe please explain why we need to include the kotlin plugin.

build.gradle
sbouchet: from GH action that check next EAP ( https://github.com/redhat-developer/intellij-openshift-connector/actions/workflows/IJ-latest.yml ) there was errors about kotlin : 
`java.lang.NoSuchMethodError: 'kotlin.sequences.Sequence kotlin.io.path.PathsKt.walk(java.nio.file.Path, kotlin.io.path.PathWalkOption[])'`
that made almost ALL tests to fails.
using that plugin made those tests passes again

sbouchet: however, looking at the differents validation builds, there is some messages i would like to check before merging this PR
```
> Task :verifyPluginConfiguration
[gradle-intellij-plugin :verifyPluginConfiguration] The following plugin configuration issues were found:
- The 'since-build' property is lower than the target IntelliJ Platform major version: 212 < 232.
- The Java configuration specifies targetCompatibility=17 but since-build='212' property requires targetCompatibility=11.
- The Kotlin configuration specifies jvmTarget=17 but since-build='212' property requires jvmTarget=11.
- The dependency on the Kotlin Standard Library (stdlib) is automatically added when using the Gradle Kotlin plugin and may conflict with the version provided with the IntelliJ Platform, see: https://jb.gg/intellij-platform-kotlin-stdlib
See: https://jb.gg/intellij-platform-versions
```

sbouchet: removed this dependency

adietish (approved):

PR #606, opened 2023-11-07
chore: extract integration tests in their own workflow

sbouchet:
fixes #???


adietish (approved):
LGTM

PR #640, opened 2023-12-05
fix: display 'plz login' also when forbidden (#631)

### LINKED ISSUE #631: When user is not logged in to any cluster, the extension still show the latest kube context and namespace field shows error
When user is logged out of a cluster, the extension still shows the latest kube context. This is not much of a problem but when the namespace field is viewed, it shows error:

Could not get namespaces: Failure executing: GET at: https://api.ocp1.adapters-crs.ccitredhat.com:6443/api/v1/namespaces. Message: Forbidden! User developer/api-ocp1-adapters-crs-ccitredhat-com:6443 doesn't have permission. namespaces is forbidden: User "system:anonymous" cannot list resource "namespaces" in API group "" at the cluster scope.

which does not feel really good.

adietish:
fixes #631


sbouchet (approved):
LGTM

PR #650, opened 2023-12-19
refactor: report telemetry when listing rels/charts

adietish:


src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
```diff
@@ -10,9 +10,12 @@
  ******************************************************************************/
 package org.jboss.tools.intellij.openshift.telemetry;
 
+import com.intellij.openapi.application.ApplicationManager;
 import com.redhat.devtools.intellij.telemetry.core.service.TelemetryMessageBuilder;
 import com.redhat.devtools.intellij.telemetry.core.util.Lazy;
 
+import static com.redhat.devtools.intellij.telemetry.core.service.TelemetryMessageBuilder.*;
```
sbouchet, line 17: please avoid using stars in imports.

adietish:
>  is the async send method should be also used in all other places ?

good point. I'll add another commit where I send async where it makes sense.

adietish:
@sbouchet replace sync send by async send where I thought it makes sense. Please re-review.

src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetrySender.java
```diff
@@ -15,6 +15,7 @@
 
 import java.util.regex.Pattern;
 
+import static org.jboss.tools.intellij.openshift.telemetry.TelemetryService.*;
```
sbouchet, line 18: please avoid star import .

src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
sbouchet: again :D

src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
```diff
@@ -61,4 +65,12 @@ public TelemetryMessageBuilder getBuilder(){
         return instance.builder.get();
     }
 
+    public static void asyncSend(ActionMessage message) {
```
sbouchet, line 68: is a usage of upperclass instance possible here ?  doesn't seems to have something specific to either ActionMessage nor FeedbackMessage.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -192,21 +194,21 @@ private void computeTelemetrySettings() {
...
         } catch (RuntimeException e) {
             // do not send telemetry when there is no context ( ie default kube URL as master URL )
             try {
                 //workaround to not send null values
                 if (e.getMessage() != null) {
-                    telemetry.error(e).send();
+                    asyncSend(telemetry.error(e));
                 } else {
-                    telemetry.error(e.toString()).send();
+                    asyncSend(telemetry.error(e));
```
sbouchet, line 211: should be 
```suggestion
                    asyncSend(telemetry.error(e.toString()));
```

adietish: @sbouchet why?

sbouchet: because above is already sending the event. this was a if/then clause to avoid sending null message in telemetry.

src/main/java/org/jboss/tools/intellij/openshift/telemetry/TelemetryService.java
adietish: not possible, unfortunately. I have made the super class/interface private :(

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
adietish: this should be fixed in telemetry then.

sbouchet: thus, no need to have this if/else anymore, and we need an issue/fix on telemetry first.

sbouchet:
star import still present after fresh checkout in commit https://github.com/redhat-developer/intellij-openshift-connector/commit/91cd835dcd60f13b1cdcd823e3895e34b8723c3d

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -192,21 +194,21 @@ private void computeTelemetrySettings() {
...
             telemetry.property(KUBERNETES_VERSION, info.getKubernetesVersion());
             telemetry.property(IS_OPENSHIFT, Boolean.toString(info.isOpenshift()));
             telemetry.property(OPENSHIFT_VERSION, info.getOpenshiftVersion());
-            telemetry.send();
+            asyncSend(telemetry);
         } catch (RuntimeException e) {
             // do not send telemetry when there is no context ( ie default kube URL as master URL )
             try {
                 //workaround to not send null values
                 if (e.getMessage() != null) {
```
sbouchet, line 208: after looking at the latest telemetry code, this if/then can be totally removed and be replaced by just  asyncSend(telemetry.error(e)); and maybe remove also the redundant try/catch

sbouchet (approved):
LGTM

PR #665, opened 2024-01-17
refactor: use stateful openshift client instead of ad hoc (#664)

### LINKED ISSUE #664: OpenShift client is not-reused, it is re-created everytime it's needed
https://github.com/redhat-developer/intellij-openshift-connector/blob/4a7c11e208923bf13106ec4eb2f79923dcdfed6c/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java#L589

https://github.com/redhat-developer/intellij-openshift-connector/blob/4a7c11e208923bf13106ec4eb2f79923dcdfed6c/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java#L834-L837

https://github.com/redhat-developer/intellij-openshift-connector/blob/4a7c11e208923bf13106ec4eb2f79923dcdfed6c/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java#L860-L861

`ClusterHelper.isOpenShift(client)` is also creating a new OpenShift client:
https://github.com/redhat-developer/intellij-common/blob/9c0aae5b4917000596cd192f2ba150c33ca893e5/src/main/java/com/redhat/devtools/intellij/common/kubernetes/ClusterHelper.java#L28
```java
  public static boolean isOpenShift(KubernetesClient client) {
        OpenShiftClient osClient = (OpenShiftClient)client.adapt(OpenShiftClient.class);
        ...
```

This is especially time consuming when it first asks whether the cluster is an OpenShift cluster (which creates an OpenShift client) and then creates an OpenShift client to retrieve data.

https://github.com/redhat-developer/intellij-openshift-connector/blob/4a7c11e208923bf13106ec4eb2f79923dcdfed6c/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java#L227-L229

adietish:
fixes #664

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
@@ -17,12 +17,12 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.deser.std.StdNodeBasedDeserializer;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.google.common.base.Strings;
 import com.intellij.execution.process.ProcessAdapter;
 import com.intellij.execution.process.ProcessEvent;
 import com.intellij.execution.process.ProcessHandler;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.util.Key;
+import com.intellij.openapi.util.text.Strings;
 import com.intellij.util.messages.MessageBusConnection;
 import com.redhat.devtools.intellij.common.kubernetes.ClusterHelper;
 import com.redhat.devtools.intellij.common.kubernetes.ClusterInfo;
@@ -52,7 +52,6 @@ import io.fabric8.openshift.client.OpenShiftClient;
 import io.fabric8.openshift.client.dsl.OpenShiftOperatorHubAPIGroupDSL;
 import io.fabric8.openshift.client.impl.OpenShiftOperatorHubAPIGroupClient;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang3.StringUtils;
 import org.jboss.tools.intellij.openshift.KubernetesLabels;
 import org.jboss.tools.intellij.openshift.utils.Serialization;
 import org.jetbrains.annotations.NotNull;
@@ -67,6 +66,7 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
+import java.net.HttpURLConnection;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -122,6 +122,7 @@ public class OdoCli implements Odo {
     private final String command;
 
     private final KubernetesClient client;
+    private final OpenShiftClient openshiftClient;
 
     private final MessageBusConnection connection;
 
@@ -153,6 +154,7 @@ public class OdoCli implements Odo {
         String current = ConfigHelper.getCurrentContextName();
         Config config = Config.autoConfigure(current);
         this.client = new KubernetesClientBuilder().withConfig(config).withHttpClientBuilderConsumer(builder -> setSslContext(builder, config)).build();
+        this.openshiftClient = toOpenShiftClient(client);
         try {
             this.envVars = NetworkUtils.buildEnvironmentVariables(this.getMasterUrl().toString());
             computeTelemetrySettings();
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -225,14 +228,11 @@ public List<String> getNamespaces() throws IOException {
     }
 
     private List<? extends HasMetadata> getNamespaces(KubernetesClient client) {
```
sbouchet, line 230: there is no need to use this param as it is the kubernetesclient instance. please remove and use directly the `client ` field

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -153,6 +155,7 @@ public OdoCli(com.intellij.openapi.project.Project project, String command) {
         String current = ConfigHelper.getCurrentContextName();
         Config config = Config.autoConfigure(current);
         this.client = new KubernetesClientBuilder().withConfig(config).withHttpClientBuilderConsumer(builder -> setSslContext(builder, config)).build();
+        this.openshiftClient = toOpenShiftClient(client);
```
sbouchet, line 158: wonder if using the client as param makes sense ? `toOpenShiftClient` method is used only at this place now.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -373,11 +373,11 @@ public List<ComponentMetadata> analyze(String path) throws IOException {
     public void createComponent(String project, String componentType, String registryName, String component, String source, String devfile, String starter) throws IOException {
         List<String> args = new ArrayList<>();
         args.add("init");
-        if (StringUtils.isNotBlank(devfile)) {
+        if (!Strings.isEmptyOrSpaces(devfile)) {
             args.add("--devfile-path");
             args.add(devfile);
         } else {
-            if (StringUtils.isNotBlank(starter)) {
+            if (!Strings.isEmptyOrSpaces(starter)) {
```
sbouchet, line 380: please do same refactor in line 655

sbouchet: and 895

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -833,15 +832,12 @@ public java.net.URL getMasterUrl() {
...
-                    VersionInfo info = oclient.getOpenShiftV3Version();
-                    if (info == null) {
-                        ConfigMap configMap = client.configMaps().inNamespace(OCP4_CONFIG_NAMESPACE).withName(OCP4_CONSOLE_PUBLIC_CONFIG_MAP_NAME).get();
-                        if (configMap != null) {
-                            return configMap.getData().get(OCP4_CONSOLE_URL_KEY_NAME);
-                        }
+            if (openshiftClient != null) {
+                VersionInfo info = openshiftClient.getOpenShiftV3Version();
+                if (info == null) {
+                    ConfigMap configMap = client.configMaps().inNamespace(OCP4_CONFIG_NAMESPACE).withName(OCP4_CONSOLE_PUBLIC_CONFIG_MAP_NAME).get();
```
sbouchet, line 838: just tested out that this code needs to use openshiftClient versus client. tested out that it actually results the same, and will improve readability.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -833,15 +832,12 @@ public java.net.URL getMasterUrl() {
...
+                    if (configMap != null) {
+                        return configMap.getData().get(OCP4_CONSOLE_URL_KEY_NAME);
                     }
                 } else {
                     ConfigMap configMap = client.configMaps().inNamespace(OCP3_CONFIG_NAMESPACE).withName(OCP3_WEBCONSOLE_CONFIG_MAP_NAME).get();
```
sbouchet, line 843: same as above

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
adietish: I would keep it. Method signature lays bare that a client is adapted, something you'd only discover reading the method body if the instance variable was used. Furthermore it makes it immediately clear that one needs to initialize the kubernetes client first.

sbouchet (approved):
LGTM

PR #668, opened 2024-01-19
chore: fix sonar workflow

### LINKED ISSUE #667: Sonar  workflow is missing reports and fails to report coverage
https://github.com/redhat-developer/intellij-openshift-connector/actions/workflows/sonar.yml is reporting no  coverage. logs indicates there is no artifacts being downloaded from inputs builds, thus no coverage can be calculated.

sbouchet:
fix sonar workflow

fixes #667

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,21 +10,6 @@ on:
     branches: [ main ]
 
 jobs:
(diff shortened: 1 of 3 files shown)


adietish (approved):
cannot verify, but looks fine to me. lgtm

PR #675, opened 2024-01-26
fix: expire 'creating project' notif if creation fails (#674)

### LINKED ISSUE #674: If I create a project which errors on the cluster, the "Creating project" notification is visible along the error dialog
**Steps:**
1. ASSERT: use the redhat sandbox
2. EXEC: Create a new project
3. ASSERT: "Creating project notification" is shown

**Result:**
image
Creating the project errors (expected. You cannot create a 2nd project on the sandbox) and the error is shown in an error dialog. The notification is still visible which makes no sense.

**Expected result:**
The notification should disappear when the error dialog appears.

adietish:
fixes #674 
depends on #666

This can only be merged once #666 is merged because it relies on changes that were introduced in it.


adietish:
replaced by #680. 
Closing.

PR #677, opened 2024-01-26
fix: offer 'Open Helm Charts' on project node (was: cluster)(#676)

### LINKED ISSUE #676: "Open Helm Charts" should be available on the project node (not the cluster) since charts are installed into the project/namespace
related to #613, #612 

"Open Helm Charts" is currently available in the context menu of the cluster node. This is misleading because charts are installed into the current project/namespace (see #613). 
image
VScode currently has it on the Helm node. 
image
We dont have a helm node yet and we should therefore offer it on the project/namespace node instead

adietish:
fixes #676 

image


sbouchet (approved):
LGTM

Issue #661, opened 2024-01-17
As a user I'd like to be able to switch current project/namespace


adietish:
related to #639, #618

Currently there's no way for me to switch the current project/namespace in the tooling. Vscode allows this via the 2 arrows to the right of the cluster in the resource tree ("Change Active Project").
<img width="686" alt="image" src="https://github.com/redhat-developer/intellij-openshift-connector/assets/25126/42ea5438-5a33-405f-bb61-708583a9a646">

PR #680, opened 2024-01-30
feat: validate new project name (#237)

### LINKED ISSUE #237: Add name validator for New project dialog
[kind/Enhancement]

Add validator for New Project dialog:
image
VSCode:
image

## Why is this needed? 
This is implemented in VSCode and JBT.

adietish:
fixes #237, #674, #701
depends on #661

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/actions/project/CreateProjectAction.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/actions/project/CreateProjectAction.java
@@ -19,11 +19,17 @@ import com.redhat.devtools.intellij.common.utils.UIHelper;
 import org.jboss.tools.intellij.openshift.actions.ActionUtils;
 import org.jboss.tools.intellij.openshift.actions.NotificationUtils;
 import org.jboss.tools.intellij.openshift.actions.cluster.LoggedInClusterAction;
+import org.jboss.tools.intellij.openshift.telemetry.TelemetryService;
 import org.jboss.tools.intellij.openshift.tree.application.ApplicationsRootNode;
+import org.jboss.tools.intellij.openshift.ui.SwingUtils;
+import org.jboss.tools.intellij.openshift.ui.project.CreateNewProjectDialog;
 import org.jboss.tools.intellij.openshift.utils.odo.Odo;
 import org.jetbrains.annotations.NotNull;
 
+import java.awt.Point;
 import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 
 import static org.jboss.tools.intellij.openshift.actions.ActionUtils.runWithProgress;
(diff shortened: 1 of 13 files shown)


sbouchet:
Same remark as #689 . is it possible to change labels if the cluster is openshift(project) or kubernetes(namespace) ?

sbouchet:
> Same remark as #689 . is it possible to change labels if the cluster is openshift(project) or kubernetes(namespace) ?

created #701 for fixing actions labels only.

adietish:
I'll fix #701 in this PR as providing different labels for kubernetes/openshift is closely related to the actions to launch it.

I investigated a unified approach where where `TreeAction#update` would set the Action label in the ctx menu. It doesn't work. Weird enough, `update` is called before the context menu is to be shown for the first time, but it has no effect. It takes only effect on the 2nd, 3rd, 4th, etc. call. 
What would have worked was to register an alternative text via `AnAction#addTextOverride` in the constructor. Unfortunately I need an `AnActionEvent` to get the root node with the odo that would tell me if we have an openshift or a kubernetes cluster (`Odo.isOpenShift()`).
This left me with the only approach where there are 2 distinct actions for openshift and kubernetes cluster that would be visible for the specified cluster type.

adietish:
@sbouchet: please re-review.

adietish:
/override ci/prow/e2e-openshift  tide "Aggregated Test Report"

src/main/java/org/jboss/tools/intellij/openshift/actions/project/CreateProjectAction.java
```diff
@@ -19,18 +19,43 @@
...
+
+  @Override
+  public void update(AnActionEvent e) {
+    super.update(e);
+    if (e.getPresentation().isVisible()) {
+      Odo odo = getOdo(e);
+      if (odo == null) {
+        return;
+      }
+      if (!odo.isOpenShift()) {
```
sbouchet, line 53: this is redundant with getKind() that does the same test.

adietish:
> there is no update on ChangeActiveProjectAction, thus the label is always 'project'.

I cannot replicate this. Things work fine for me when switching from kind to crc.

https://github.com/redhat-developer/intellij-openshift-connector/assets/25126/63958103-7872-4d92-ab37-40560b999e58


What am I missing?

sbouchet:
> @sbouchet:
> 
> > there is no update on ChangeActiveProjectAction, thus the label is always 'project'.
> 
> I cannot replicate your issue. Things are working fine for me when switching from kind to crc.
> 
>  Screen.Recording.2024-02-08.at.20.17.39.mov 
> What am I missing?

look at 0:13. this is not the "create" label, but the "Change " one that is always displaying "Project". there is a missing update() method in the ChangeAction.

adietish:
> > @sbouchet:
> > > there is no update on ChangeActiveProjectAction, thus the label is always 'project'.
> > 
> > 
> > I cannot replicate your issue. Things are working fine for me when switching from kind to crc.
> > Screen.Recording.2024-02-08.at.20.17.39.mov
> > What am I missing?
> 
> look at 0:13. this is not the "create" label, but the "Change " one that is always displaying "Project". there is a missing update() method in the ChangeAction.

This PR is about the "New Project" wizard, not the "Change Project" wizard :)
So this is a different issue, I'll file and fix it.

adietish:
@sbouchet 👍 gotcha, this was included in your PR for #701. Adding it here, too.

adietish:
@sbouchet please re-review, thanks!

sbouchet:
Capture d’écran du 2024-02-09 12-12-47
left two places with "project" when using kubernetes cluster. 
- one on validation message on dialog 
- one on missing project/namespace message on tree

adietish:
@sbouchet fixes those 2, please re-review.

adietish:
/override ci/prow/e2e-openshift "Aggregated Test Report" tide

PR #683, opened 2024-01-30
fix: report 'cluster unreachable' instead of 'failed to list secrets' (#682)

### LINKED ISSUE #682: If cluster is down, resource tree reports a failing operation to list secrets
**Steps:**
1. ASSERT: have the current context (in kube config) pointing to a cluster that is down
2. EXEC: launch plugin

**Result:**
The child node to the cluster node reports failing to list secrets.
image

**Expected result:**
It should report that the cluster is down.

adietish:
fixes #682


sbouchet (approved):
LGTM. nice to see new coverage, too !

adietish:
/override "Conventional Commits PR Check / Aggregated Test Report (pull_request_target)"

PR #697, opened 2024-02-02
fix: show 'no url/binding' node, use async leafstate for component (#687)

### LINKED ISSUE #687: URL aren't displayed anymore when component is on dev mode
[kind/bug]

- Operating System: Fedora
- IntelliJ version: 2023.3
- OpenShift version: 

## What did you run exactly?
When running component in dev mode, URLS are'nt displayed anymore as child of the component.

## Actual behavior
current main 

Capture d’écran du 2024-02-01 11-25-54

## Expected behavior
from 1.6.0

Capture d’écran du 2024-02-01 11-21-01

adietish:
fixes #687

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
@@ -95,9 +95,7 @@ public class ApplicationsTreeStructure extends AbstractTreeStructure implements 
 
     @Override
     public @NotNull LeafState getLeafState(@NotNull Object element) {
-        if (element instanceof ComponentNode) {
-            return LeafState.ALWAYS;
-        } else if (element instanceof ChartReleaseNode) {
+        if (element instanceof ChartReleaseNode) {
             return LeafState.ALWAYS;
         } else if (element instanceof MessageNode<?>) {
             return LeafState.ALWAYS;
@@ -233,10 +231,12 @@ public class ApplicationsTreeStructure extends AbstractTreeStructure implements 
             return Collections.emptyList();
         }
         try {
-            odo.listURLs(element.getParent().getName(),
-                element.getComponent().getPath(), element.getName()).forEach(url ->
-                    results.add(new URLNode(element, url))
-                );
+            odo.listURLs(
+                element.getParent().getName(),
+                element.getComponent().getPath(), element.getName())
+              .forEach(url ->
+                results.add(new URLNode(element, url))
+              );
         } catch (IOException e) {
             LOGGER.warn(e.getLocalizedMessage(), e);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
```diff
@@ -94,9 +94,7 @@ public Object getApplicationsRoot() {
 
     @Override
     public @NotNull LeafState getLeafState(@NotNull Object element) {
-        if (element instanceof ComponentNode) {
-            return LeafState.ALWAYS;
```
adietish, line 98: this is the fix: the leafstate for `ComponentNode` is now async

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
```diff
@@ -221,34 +219,47 @@ private List<BaseNode<?>> load(Callable<List<BaseNode<?>>> callable, NamespaceNo
...
+                componentNode.getParent().getName(),
+                componentNode.getComponent().getPath(),
+                componentNode.getName())
+              .stream()
+              .map(url -> new URLNode(componentNode, url))
+              .collect(Collectors.toList());
+            if (!urls.isEmpty()) {
+                nodes.addAll(urls);
+            } else {
+                nodes.add(new MessageNode<>(root, componentNode, "<No URLS>"));
```
adietish, line 236: @sbouchet: suggested enhancement: add a node "no URLs" if no urls were found for the component

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
```diff
@@ -221,34 +219,47 @@ private List<BaseNode<?>> load(Callable<List<BaseNode<?>>> callable, NamespaceNo
...
+                componentNode.getParent().getName(),
+                componentNode.getComponent().getPath(),
+                componentNode.getName())
+              .stream()
+              .map(binding -> new BindingNode(componentNode, binding))
+              .collect(Collectors.toList());
+            if (!bindings.isEmpty()) {
+                nodes.addAll(bindings);
+            } else {
+                nodes.add(new MessageNode<>(root, componentNode, "<No Bindings>"));
```
adietish, line 257: @sbouchet: suggested enhancement: add a child "no bindings" if no urls were found for the component

src/main/java/org/jboss/tools/intellij/openshift/tree/application/ApplicationsTreeStructure.java
```diff
@@ -318,6 +329,11 @@ public Object getParentElement(@NotNull Object element) {
     public void commit() {
     }
 
+    @Override
+    public boolean isToBuildChildrenInBackground(@NotNull Object element) {
+        return true;
```
adietish, line 334: enhancement: retrieve children in a background thread (instead of UI thread)

sbouchet:
@adietish : your enhancement are good idea, but may need more than 'just' displaying no url/url.
letme explain. 
URL are actually shown only when running on dev mode, because they are 'active' and the context action opens a webbrowser to that address.
If your idea is to show 'no URL', it shouldbe clear taht it's no ACTIVE url that can be browsed, because when you describe the component using odo, it always shows URL taht are listed on the devfile.
What i propose is to display more information.
- NO URL only when there is no URL in the devfile
- URL in grey state, when there is URL in the devfile but the compoent is not running
- URL in White/Green state, when the URL is active

this is somehow similar to what odo does, and we can rely on it to parse the state ( try out with descibe commands) .

WDYT ?

sbouchet:
thanks for filling a new enhancement request.

sbouchet (approved):
LGTM. please follow up with #707

PR #702, opened 2024-02-05
feat: Use "namespace" when using kubernetes cluster

### LINKED ISSUE #701: Fix Action labels for kubernetes clusters
[kind/Enhancement]

When using a kubernetes-based cluster, i would like to see "Create Namespace/Change namespace" and not Openshift "project" labels in context menus.

sbouchet:
use "namespace" when using kubernetes cluster

fixes #701 
need #680


sbouchet:
closing as fixed with #680

PR #703, opened 2024-02-05
chore: Remove unnecessary files

sbouchet:
remove not needed files


adietish (approved):
image
This icon indeed isn't used anywhere in the plugin.

sbouchet:
> image
> This icon indeed isn't used anywhere in the plugin. We should maybe use it as cluster icon though to match what vscode is using.

@adietich there is two icons, one in  the images directory (openshift_extension.png) and oone in the META-INF (openshiftpluginicon13*13.png ) . the later is the one used for the view and must remain, the other one in the feedback dialog. 

which one do you refer to ?

PR #725, opened 2024-02-28
chore: refactor workflows and issues template

sbouchet:
fixes #???


.github/workflows/ui_it.yml
```diff
@@ -0,0 +1,56 @@
+name: UI and Integration Tests
+on:
+  workflow_run:
+    workflows: [ "Java CI with Gradle" ]
```
adietish, line 4: @sbouchet: I most likely dont get this fully: 
this workflow should be triggered when the workflow `Java CI with Gradle` is completed. But there isn't a workflow called `Java CI with Gradle`, isn't it?

adietish (approved):
assuming that I didnt fully understand it, LGTM

.github/workflows/ui_it.yml
sbouchet: it's https://github.com/redhat-developer/intellij-openshift-connector/blob/main/.github/workflows/ci.yml

PR #773, opened 2024-04-04
fix: start dev on podman does not work if odo can't retrieve podman status

### LINKED ISSUE #637: Start dev on Podman does not work when odo fails to fetch version
When I try to start dev on Podman I get an error message:
`unable to access podman. Do you have podman client installed and configured correctly? cause: exec: "podman": executable file not found in $PATH`
However, if I run the same command from terminal, starting dev works well.

I am running intellij 2023.2.4 (Ultimate Edition) on MacOS 14.1.1

sbouchet:
fixes #637


src/main/java/org/jboss/tools/intellij/openshift/actions/component/DevOnPodmanComponentAction.java
```diff
@@ -33,4 +42,25 @@ protected String getCustomizedPresentation(Component component) {
...
+            if (!(node instanceof ComponentNode)) {
+                return;
+            }
+            ComponentNode componentNode = (ComponentNode) node;
+            Odo odo = componentNode.getRoot().getOdo().getNow(null);
+            if (odo == null) {
+                return;
+            }
+            try {
+                e.getPresentation().setEnabled(odo.checkPodman());
```
sbouchet, line 59: Actual fix is here

adietish (approved):
works for me. LGTM. 
I would have preferred to have the reformatting in a separate PR though because it's almost impossible to review 4000+ lines.

adietish (changes requested):
Imho we should fix the fatal IDE error that I can reproduce in `IU-2023.3`

src/main/java/org/jboss/tools/intellij/openshift/actions/component/DevOnPodmanComponentAction.java
sbouchet: nope

sbouchet:
> works for me. LGTM. I would have preferred to have the reformatting in a separate PR though because it's almost impossible to review 4000+ lines.

Yes, totally agree. as tests are failling, i'm going to a rabbit hole trying to refactor it rather than fixing it and focus on the actual isssue.

src/main/java/org/jboss/tools/intellij/openshift/actions/component/FeatureComponentAction.java
```diff
@@ -59,10 +59,6 @@ public void update(AnActionEvent e) {
             }
             ComponentNode componentNode = (ComponentNode) node;
             Component component = componentNode.getComponent();
-            Odo odo = componentNode.getRoot().getOdo().getNow(null);
```
sbouchet, line 62: removed because it causes slow actions updates.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -459,16 +459,18 @@ public List<URL> listURLs(String context) throws IOException {
   @Override
   public ComponentInfo getComponentInfo(String project, String component, String path,
                                         ComponentKind kind) throws IOException {
+    boolean isPodmanPresent = !execute(command, envVars, "version").contains("unable to fetch the podman client version");
```
sbouchet, line 462: podman check is now here, added as componentInfo status. This is done when loading tree component where the debug podman action resisdes.

sbouchet:
@adietish please review.

src/test/java/org/jboss/tools/intellij/openshift/utils/odo/JsonParserTest.java
```diff
@@ -87,26 +87,26 @@ public void verifyThatURLSCanBeLoadedFromDeployModeOnOpenShift() throws IOExcept
...
+  @Test
+  public void verifyThatComponentInfoCanBeLoadedFromDescribe() throws IOException {
+    URL url = JsonParserTest.class.getResource("/describe-component-local.json");
+    JSonParser parser = new JSonParser(MAPPER.readTree(url));
+    ComponentInfo componentInfo = parser.parseDescribeComponentInfo(ComponentKind.DEVFILE, false);
+    assertNotNull(componentInfo);
+    assertTrue(componentInfo.getSupportedFeatures().contains(ComponentFeature.Mode.DEV_MODE));
+    assertEquals("Go", componentInfo.getComponentTypeName());
+    assertEquals("Go", componentInfo.getLanguage());
+  }
```
adietish, line 111: Maybe you could add a unit test in `DevOnPodmanComponentActionTest` testing enablement when podman is present/not present?

adietish (approved):
Other than maybe adding a unit test for the action, LGTM. Works for me.

PR #805, opened 2024-04-24
fix: fix Integration tests

sbouchet:
current main integration tests are failing.


adietish (approved):
LGTM. With all the reformatting it is hard for me to spot what exactly fixes the ITs. Can you please point out?

sbouchet:
> LGTM. With all the reformatting it is hard for me to spot what exactly fixes the ITs. Can you please point out?

well, it's only the usage of a new test project, from springboot to nodejs. 
i also added a devfile 2.2.0 along with the old 2.0.0 as separate unit test for regression.

PR #821, opened 2024-05-15
build(deps): bump helm/kind-action from 1.9.0 to 1.10.0 part two 

sbouchet:
fixes integrationUITests


sbouchet:
see #810 for part 1

sbouchet:
LGTM

PR #846, opened 2024-06-21
refactor: Refactoring Odo Interface and integration Tests

### LINKED ISSUE #779: Refactor integration tests
from #773
```
  >> works for me. LGTM. I would have preferred to have the reformatting in a separate PR though because it's almost impossible to review 4000+ lines.

  >Yes, totally agree. as tests are failing, i'm going to a rabbit hole trying to refactor it rather than fixing it and focus on the actual issue.
```

### LINKED ISSUE #820: Refactor Odo interfaces to avoid 'UnsupportedOperationException' in Implementation classes
remove all 'UnsupportedOperationException' in classes

sbouchet:
fixes #779 
fixes #820

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/Odo.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/Odo.java
@@ -11,110 +11,81 @@
 package org.jboss.tools.intellij.openshift.utils.odo;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.intellij.execution.process.ProcessAdapter;
-import com.intellij.execution.process.ProcessHandler;
 
 import java.io.IOException;
 import java.util.List;
-import java.util.function.Consumer;
-
-import static org.jboss.tools.intellij.openshift.Constants.DebugStatus;
 
 public interface Odo {
-    List<String> getNamespaces() throws IOException;
+  List<String> getNamespaces() throws IOException;
 
-    String getCurrentNamespace();
+  String getCurrentNamespace();
 
-    boolean namespaceExists(String name);
+  boolean namespaceExists(String name);
 
-    String getNamespaceKind();
+  String getNamespaceKind();
 
-    void start(String context, ComponentFeature feature, ProcessHandler handler, ProcessAdapter processAdapter) throws IOException;
+  void describeComponent(String context) throws IOException;
 
-    void start(String context, String component, ComponentFeature feature,
-               Consumer<Boolean> callback, Consumer<Boolean> processTerminatedCallback) throws IOException;
+  List<ComponentMetadata> analyze(String path) throws IOException;
 
-    void stop(String context, String component, ComponentFeature feature) throws IOException;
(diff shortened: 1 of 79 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/Odo.java
```diff
@@ -11,110 +11,83 @@
...
-    List<DevfileComponentType> getComponentTypes(String name) throws IOException;
+  void deleteDevfileRegistry(String name) throws IOException;
 
-    boolean isOpenShift();
+  boolean isOpenShift();
 
-    void migrateComponent(String name);
+  void migrateComponent(String name);
 
+  boolean checkPodman() throws IOException;
```
adietish, line 92: Wouldn't a more speaking name be `canExecutePodman()`? Unless I am misreading the purpose of `OdoCli.checkPodman()`

src/test/java/org/jboss/tools/intellij/openshift/actions/project/CreateProjectActionTest.java
```diff
@@ -11,46 +11,41 @@
...
   @Override
-  protected void verifyLoggedInCluster(boolean visible) {
-    assertTrue(visible);
+  protected void verifyLoggedInCluster(@NotNull Presentation presentation) {
+    assertTrue(presentation.isVisible());
+    assertEquals("New " + (isOpenshift ? "Project" : "Namespace"), presentation.getText());
   }
-
+/*
   public void testActionOnProject() {
```
adietish, line 34: Can you please explain why you're commenting these tests?

adietish (commented):
Big kudos for the introduction of these 2 interfaces and getting rid of the methods that threw `UnsupportedOperationException`. Feels much more clean now.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/Odo.java
sbouchet: good review. i'll change that.

src/test/java/org/jboss/tools/intellij/openshift/actions/project/CreateProjectActionTest.java
sbouchet: leftovers to be removed.

src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoDelegate.java
```diff
@@ -0,0 +1,21 @@
+package org.jboss.tools.intellij.openshift.utils.odo;
```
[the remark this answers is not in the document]
sbouchet: You're correct with all the above assumptions :smile: 
`OdoFacade` is for consumers, like the resource tree and most of the tests. 
`OdoDelegate` is for back end implementors, like the CLI.
why we need two is because there is a intermediate Helper that holds the current running odo processes.

sbouchet:
@adietish please re-review after your comments

PR #851, opened 2024-06-25
fix: Update helm to 3.14.4

sbouchet:
See https://mirror.openshift.com/pub/openshift-v4/clients/helm


adietish (approved):
LGTM. Downloaded all binaries, checksum'ed them and tested it on MacOS while removing my 3.13 helm binary beforehand. Works for me.

PR #852, opened 2024-06-26
fix: Update odo to 3.16.1

sbouchet:
See https://github.com/redhat-developer/odo/releases/tag/v3.16.1


adietish (approved):
LGTM. Verified all urls and tested it on MacOS killing my odo cache before hand. Works for me.

PR #866, opened 2024-07-05
fix: unable to login on cluster with private certificate

### LINKED ISSUE #862: Unable to login: failed to verify certificate
- Operating System: Fedora
- IntelliJ version: IntelliJ IDEA 2023.3.4 (Ultimate Edition)
- Extension version: 1.9.1-20240702-013100
- Are you targeting a kubernetes or an Openshift cluster ? If possible, provide its version:  https://api.ocp2.adapters-crs.ccitredhat.com:6443

## What did you run exactly?

- Log in to the cluster using the extension
image

## Actual behavior

image

mszuc@fedora:~/.odo/cache/3.16.1$ ./odo login -p x -u x https://api.ocp2.adapters-crs.ccitredhat.com:6443
Connecting to the OpenShift cluster

 x  tls: failed to verify certificate: x509: certificate signed by unknown authority

sbouchet:
Fix and migrate login action to use OC

fixes #862


src/main/java/org/jboss/tools/intellij/openshift/actions/cluster/LoginAction.java
```diff
@@ -37,18 +37,18 @@ public LoginAction() {
   public String getTelemetryActionName() {return "login to cluster";}
 
   @Override
-  public void actionPerformedOnSelectedObject(AnActionEvent anActionEvent, Object selected, @NotNull OdoFacade odo) {
+  public void actionPerformedOnSelectedObject(AnActionEvent anActionEvent, Object selected, @NotNull Oc oc) {
```
adietish, line 40: I found a problem with the `LoginDialog` but I dont know where the problem is exactly:

**Steps:**
1. ASSERT: make sure that you dont have oc in the cache:
`rm -rf ~/.oc/cache`
2. EXEC: pick "Log in to Cluster" in the ctx menu of the cluster node
3. ASSERT: you see "Downloading oc" with a progress bar in the bottom right corner of the IDE window

**Result:**
Nothing happens, the Login dialog does not show up. You have to pick again "Log in to Cluster" in the context menu.

sbouchet: can reproduce this. investigating

sbouchet: ok, debugged and turns out the first call on login returns a null OcTool as completable future is not complete, so the login doesn't shows up. in the background the tool is downloaded, then in the second call, the tool is correclty returned. will modify the logic.

adietish (approved):
great work, works for me. LGTM

Issue #808, opened 2024-04-24
I want to be able to "Refresh" the helm repositories


adietish:
**Steps:**
1. EXEC: expand the resource tree [cluster] > Helm Repositories so that you see the helm repos
<img width="480" alt="image" src="https://github.com/redhat-developer/intellij-openshift-connector/assets/25126/a0012c9b-3bc4-4d04-b435-e7326644c4cc">

2. EXEC: execute in cli: `helm repo add gerkelznik https://gerkelznik.github.io/helm-charts`
3. ASSERT: the new helm repo `gerkelznik` is not listed
4. EXEC: now try to refresh the list of repositories

**Result:**
There's no "Refresh" in the context menu for the "Helm Repositories" node. You can refresh the list by refreshing the cluster though by using "Refresh" in the context menu to the cluster node

sbouchet:
While i'm ok with this feature, next question is why we need to have the Helm node located under the cluster node.
Can it be next to it, like the devfile registries ? is there an hard link between the Helm node and the CLuster one ?

PR #867, opened 2024-07-05
test: added test for RefreshAction for helm repositories (#808)

adietish:


sbouchet (approved):
LGTM. always nice to have test suits matching features

PR #872, opened 2024-07-09
fix: removed usage of PopupHandlerAdapter

adietish:
we dont need the PopupHandlerAdapter any more since we moved >`IC-2022.1`


sbouchet (approved):
LGTM

PR #893, opened 2024-08-07
fix(UItest): extended timeout for login dialog for test stability

### LINKED ISSUE #892: Public Integration Test - Login Dialog tests unstable
Tests are unstable and sometimes fail because of a too short timeout for opening the Login to Cluster dialog

martinszuc:
Fixes #892. With my apologies I have noticed that the public tests have an instability to them, where every third or so would fail because of context menu or Login Dialog itself. I ran the test with this fix 20 or more times to verify and this fixes the instability.
image

### DIFF
--- a/.github/workflows/cluster_integration_ui_tests.yml
+++ b/.github/workflows/cluster_integration_ui_tests.yml
@@ -49,6 +49,7 @@ jobs:
           sleep 10
           ./gradlew clusterIntegrationUITest --continue --no-daemon --info
       - name: Publish tests reports
+        if: always()
         uses: scacap/action-surefire-report@a2911bd1a4412ec18dde2d93b1758b3e56d2a880 #v1.8.0
         with:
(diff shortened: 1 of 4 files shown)


sbouchet:
I may also add a a change on the workflows to always publish reports even on failures. right now if the test fails like https://github.com/redhat-developer/intellij-openshift-connector/actions/runs/10286593883/job/28467630163, no test reports is produced and then the check is waiting for it (Public Integration UI Tests Report Expected — Waiting for status to be reported)

sbouchet (changes requested):
please add on both Cluster and public IT test a 
`if: always() ` on the publish report task here https://github.com/redhat-developer/intellij-openshift-connector/blob/9d1e9ee0bff290a86b48a9748f6eedac880307d5/.github/workflows/public_integration_ui_tests.yml#L36

sbouchet (approved):
LGTM 
let's hope this will stabilize our integration tests suite

PR #911, opened 2024-09-17
chore(skip-release): bump intellij common 1.9.6

adietish:
bumping intellij-common to `1.9.6(.Final)` because `1.9.6-SNAPSHOT` isn't available from repo any more.


adietish:
@sbouchet: what's happening here? `Conventional Commits PR Check / Cluster Integration UI Tests Report (pull_request_target)` fails but there are no results for it? Do you know?

PR #915, opened 2024-09-18
fix: update tree, query cluster info and send telemetry in backgroud (#912)

### LINKED ISSUE #912: When launching the plugin on an existing project, I cannot use it. IndicatorCancellationException
**Steps:**
1. ASSERT: use a project that is already indexed.
1. ASSERT: have RH Sandbox as current cluster
1. EXEC: launch the plugin

**Result:**
The resource tree shows an error and thus most actions are not available in the context menus:

ps. I am still not fully sure how to/not to reproduce it. It happens always to me.

Workaround for me: I touch `~/.kube/config` which causes the resource tree to be refreshed and display the cluster, current namespace etc.

adietish:
fixes #912

### DIFF
--- a/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
+++ b/src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
@@ -20,9 +20,7 @@ import com.intellij.execution.process.ProcessHandler;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.util.text.Strings;
 import com.intellij.util.messages.MessageBus;
-import com.intellij.util.messages.MessageBusConnection;
 import com.redhat.devtools.intellij.common.utils.ExecHelper;
-import com.redhat.devtools.intellij.telemetry.core.configuration.TelemetryConfiguration;
 import io.fabric8.kubernetes.api.Pluralize;
 import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.DeletionPropagation;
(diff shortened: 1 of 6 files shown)


src/main/java/org/jboss/tools/intellij/openshift/utils/odo/OdoCli.java
```diff
@@ -418,7 +415,7 @@ public ComponentInfo getComponentInfo(String project, String component, String p
 
   private ComponentInfo parseComponentInfo(String json, ComponentKind kind) throws IOException {
     JSonParser parser = new JSonParser(Serialization.json().readTree(json));
-    return parser.parseDescribeComponentInfo(kind, isPodmanPresent);
+    return parser.parseDescribeComponentInfo(kind, isPodmanPresent());
```
sbouchet, line 418: why modify this variable to a method call each time a compoenent is parsed ? this is time consuming for no reason, has podman is present or not for the whole time the extension is instanciated.

adietish: the reason is the following: the check was executed in the constructor and it's a heavy operation. All the heavy operations that are in the constructor of OdoCli get executed when the tree structure requests the odo binary. This then lead to the tree displaying "IndicatorCancelled" when the UI was waiting for too long for these operations to be finished.
In this particular case, the instance variable isn't even needed. There's only 1 place where it's value is needed. That's where I replaced the instance var by a direct call.

sbouchet (approved):
LGTM. heavy check is now perfomed only once, in a lazy manner. good one !

PR #919, opened 2024-09-24
fix: remove open console action.

### LINKED ISSUE #567: Move open dashboard action to kubernetes
[kind/Enhancement]

when https://github.com/redhat-developer/intellij-kubernetes/issues/617 will be merged, our own open console will be a duplicate.

sbouchet:
fixes #567


adietish:
@sbouchet: current failures in checks are caused because checks are run against main, correct?

adietish (approved):
lgtm. I ran test locally on the branch and they are fine.

sbouchet:
@adietish Cluster UI integration tests are failling since a while. with no obvious reasons. 
i agree that this is not ideal but we can't use the fork as running branch, as we use a gpg secret key to use IDEA Ultimate.

adietish:
/override Conventional Commits PR Check / Cluster Integration UI Tests Report

adietish:
/override Conventional Commits PR Check / Cluster Integration UI Tests Report (pull_request_target)

PR #920, opened 2024-09-27
fix: replaced helm png by svg, use resized, not scaled icon (#874)

### LINKED ISSUE #874: Helm chart icons increase in size when highlighted with New UI
- IntelliJ version: IntelliJ IDEA 2023.3.7 (Ultimate Edition)

## What did you run exactly?

In Intellij enable New UI (Settings/Appearance & Behavior/New UI/Enable new UI)

[Screencast from 2024-07-15 11-57-07.webm](https://github.com/user-attachments/assets/15c62887-ba19-4033-a61c-3bbd1a164f95)

Helm icons change in size when highlighted. I have not encountered this with New UI disabled.

adietish:
fixes #874


adietish:
@sbouchet:
The icon used here has a size of 32x32. When using it for the tree node, it gets scaled to 16x16 before handing it over. Text and Icons are provided to a node in a so called "Presentatin". When a node is selected, the presentation is updated. Weird enough, the updated presentation for a selected node is not respecting the scaled size of the icon it was given. It is using the unscale/original size of it.
I thus provide 2 different icons: a 32x32 variant and a 16x16 variant. The 16x16 variant is to be used in the tree, the 32x32 is to be used in the helm charts dialog.
Please review. Thx.

sbouchet:
there is still a wrong size for installed charts, when selected :
image

adietish:
@sbouchet: I suspec that this was caused by a bug in the IDEA platform: we were using "scaled" icons (16px in the tree, 32px in the charts dialog) - using `SwingUtils#scaleIcon`. Those keep their original size but indicate a resized display size. The tree renderer seems not to use the scaled size when rendering selected nodes while it does when those are not selected. There's not such problem when using an icon with a scaled size, not just a display size - using `IconUtil#resizeSquared`

adietish:
@sbouchet: please re-review

sbouchet (approved):
tested against new UI and Old UI, everything is correct !

